PR #124, opened 2019-07-20
Modify  json's date to Json standard time format ISO-8601

maikebing:
Reason:
If it's not a standard format  then requires  conversion, it's very unnecessary.
References:
https://www.w3.org/TR/NOTE-datetime
https://en.wikipedia.org/wiki/ISO_8601
https://stackoverflow.com/questions/10286204/the-right-json-date-format

### DIFF
--- a/src/modules/http/src/httpJson.c
+++ b/src/modules/http/src/httpJson.c
@@ -240,11 +240,11 @@ void httpJsonTimestamp(JsonBuf* buf, int64_t t) {
   struct tm* ptm;
   time_t     tt = t / 1000;
   ptm = localtime(&tt);
-  int length = (int)strftime(ts, 30, "%Y-%m-%d %H:%M:%S", ptm);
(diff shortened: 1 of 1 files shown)


guanshengliang:
Thanks for opening this issue
I will handle it as soon as possible
For I need to extract the time zone number as accurate as possible

guanshengliang:
Restful provides three interfaces：
1. `restful/sql` will return local timestring
2. `restful/sqlt` will return timestmap as int64
3. `restful/sqlutc` will return utc timestaring

for examples, if a timestamp is `1565450243537`
1. `restful/sql` will return `2019-08-10 23:06:30.562`
2. `restful/sql` will return `1565450243537`
3. `restful/sqlutc` will return `2019-08-10T23:06:30.562+0800`

guanshengliang:
Handling a bug in microsecond precision

PR #177, opened 2019-07-24
fix #174

### LINKED ISSUE #174: memcpy overlap bug
version 1.6.1.0, tscParseInsert.c, start from line 716
```
if (cstart != NULL && createTable == true) {
    /* move the column list to start position of the next accessed points */
    memcpy(sql - len, cstart, len);
    *sqlstr = sql - len;
```
`memcpy(sql-len, cstart, len)`, sql - len 与 cstart ~ cend 之间的地址有重叠，导致错误

xwangg:
here we need memmove instead of memcpy, because destination and source may(very likely to) be overlapped


guanshengliang:
Thanks for create this pull request

PR #195, opened 2019-07-26
minor improvements to 'setErrMsg'

localvar:
1. code like `char msg[] = "xxxxxx"` unnecessarily copy the string to stack, changed to `const char* msg = "xxxxxx"`.
2. the 2rd argument of `setErrMsg` is changed from `char*` to `const char*`.
3. the 3rd argument of `setErrMsg` is removed as `strncpy` stops copying at the end of source string.


guanshengliang:
Thank you for your contribution, we have already review and accept it.

PR #332, opened 2019-08-12
fix dead code and enhance scheduler and memory pool

localvar:
first commit:
1. fix dead code.

second commit: enhance robustness of scheduler
1. check all memory allocation failure in `taosInitScheduler`;
2. make `pSched->numOfThreads` the actual number of created threads
to avoid crash in `taosCleanUpScheduler` when `pSched->qthread` is NULL
and other issues;
3. check interruption of `sem_wait` in `taosScheduleTask`,
and use `pTrace` instead of `pError` when `sem_wait` was interrupted;
4. cancel all threads before join them to enable cocurrent cancellation.

third commit: improve memory pool
1. fix memory leak:
`memset(&pool_p, 0, sizeof(pool_p))` sets the pointer to NULL,
so `free(pool_p)` does nothing.
2. improve performance by move some code to the outside of the lock.

### DIFF
--- a/src/util/src/tmempool.c
+++ b/src/util/src/tmempool.c
@@ -64,6 +64,7 @@ mpool_h taosMemPoolInit(int numOfBlock, int blockSize) {
 
   pthread_mutex_init(&(pool_p->mutex), NULL);
 
+  memset(pool_p->pool, 0, (size_t)(blockSize * numOfBlock));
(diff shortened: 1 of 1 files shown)


localvar:
close this one as single PR multiple patches  is a bad practice,  will create seperate ones later.

PR #339, opened 2019-08-13
temporary fix for running server-side application on WSL

justinpan0:
This is a temporary fix for running server-side service on Windows Subsystem for Linux (WSLv1) all versions of Ubuntu.

The original problem of `SO_NO_CHECK` is likely associated with [this WSL issue](https://github.com/microsoft/WSL/issues/3873)

**As for a temporary fix, check if the Linux is running WSL. If yes, skip the `SO_NO_CHECK` step.** 

> Note that `SO_NO_CHECK` supposedly disables Linux kernel to calculate checksum value of all packets received from TDEngine UDP server application. Disabling it might inducing extra error messages, yet this is a quick workaround especially WSL has become one of the most popular platforms thus far.
> 
> Also note that there are more issues involving /proc/[pid]/io and taosGetBandSpeed(), and I am actively investigating them as well.

### DIFF
--- a/src/util/src/tsocket.c
+++ b/src/util/src/tsocket.c
@@ -23,14 +23,17 @@
 #include <pthread.h>
 #include <stdarg.h>
 #include <sys/types.h>
+#include <sys/utsname.h>
 
 #include "os.h"
 #include "tglobalcfg.h"
 #include "tlog.h"
 #include "tsocket.h"
 #include "tutil.h"
+#include "tsystem.h"
 
 unsigned int ip2uint(const char *const ip_addr);
+bool isRunningWSLv1();
 
 /*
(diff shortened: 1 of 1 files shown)


Scythemen:
Hit the same error. 

Start to install TDEngine...
sed: can't read /etc/inittab: No such file or directory
Couldn't find an alternative telinit implementation to spawn.
cp: cannot stat '%{init_d_dir}/taosd': No such file or directory


ENV: 
Windows 10 21H2, WSL2, debian 9
TDengine 2.4.x

sangshuduo:
@Scythemen we did not officially support WSL due to the systemd lacks. I know someone is working in progress to make systemd on WSL. Right now, you can launch taosd manually as wrokaroud.

PR #344, opened 2019-08-14
enhance robustness of scheduler

localvar:
1. check all memory allocation failure in `taosInitScheduler`;
2. make `pSched->numOfThreads` the actual number of created threads
to avoid crash in `taosCleanUpScheduler` when `pSched->qthread` is NULL
and other issues;
3. check interruption of `sem_wait` in `taosScheduleTask`,
and use `pTrace` instead of `pError` when `sem_wait` was interrupted;
4. cancel all threads before join them to enable cocurrent cancellation;
5. remove unused global variable.

### DIFF
--- a/src/util/src/tsched.c
+++ b/src/util/src/tsched.c
@@ -38,17 +38,19 @@ typedef struct {
   SSchedMsg *     queue;
 } SSchedQueue;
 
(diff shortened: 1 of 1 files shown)


jtao1735:
thanks for making this piece of code more robust.

PR #345, opened 2019-08-14
improve memory pool

localvar:
1. fix memory leak:
`memset(&pool_p, 0, sizeof(pool_p))` sets the pointer to NULL,
so `free(pool_p)` does nothing.
2. improve performance by move some code to the outside of the lock.

### DIFF
--- a/src/util/src/tmempool.c
+++ b/src/util/src/tmempool.c
@@ -64,6 +64,7 @@ mpool_h taosMemPoolInit(int numOfBlock, int blockSize) {
 
   pthread_mutex_init(&(pool_p->mutex), NULL);
 
+  memset(pool_p->pool, 0, (size_t)(blockSize * numOfBlock));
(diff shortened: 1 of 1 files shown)


src/util/src/tmempool.c
```diff
@@ -125,6 +126,5 @@ void taosMemPoolCleanUp(mpool_h handle) {
   pthread_mutex_destroy(&pool_p->mutex);
   if (pool_p->pool) free(pool_p->pool);
   if (pool_p->freeList) free(pool_p->freeList);
-  memset(&pool_p, 0, sizeof(pool_p));
```
jtao1735, line 128: this line shall be kept, it is a good practice, since app may still access this piece of memory.

localvar: revised according to comment, please check again.

PR #383, opened 2019-08-18
fix & enhancement for udp connection handling

localvar:
1. potential data race in `taosProcessMonitorTimer`:
the issue does not exist at present because there's only one scheduler
thread, which means there's no cocurrent calls to this function for a
same `pMonitor`. but if more scheduler threads are created later,
there's a data race issue in rare case. as threads number can be
easily increased by increase the value of `taosTmrThreads`, it is very
unlikely that the developer could realize this function need to be
revised together. that's why i say it is a 'potential' issue.

this issue happens in below scenario:
a. scheduler thread1: `if (pSet)` is true and new timer is installed by
`taosTmrReset`;
b. scheduler thread1: `if (pMonitor->pSet == NULL)` is true but
`taosTmrStopA` is blocked, either by the mutex of the timer or os thread
scheduler.
c. timer thread: 200ms elapse, new call to `taosProcessMonitorTimer` is
initialized in scheduler thread2;
d. scheduler thread2: `if (pMonitor->pTimer != tmrId)` is false;
e. scheduler thread1: unblocked, stops timer and frees `pMonitor`;
f. scheduler thread2: unexpected behavior because `pMonitor` is not
valid any more.

because the result of this issue is crash or worse, i suggest to fix it
though the possibility of all the conditions are met is very very low.

2. `pthread_attr_t` related issues: per manual, an initialized
`pthread_attr_t` can be reused, should be destroyed, and the behavior of
re-init it is undefined. this issue exist in other modules also.

3. memory leaks;

4. improve failure case handling of `taosInitUdpConnection`;

5. typo

### DIFF
--- a/src/rpc/src/tudp.c
+++ b/src/rpc/src/tudp.c
@@ -137,9 +137,7 @@ void taosProcessMonitorTimer(void *param, void *tmrId) {
     tTrace("%s monitor timer is expired, update the link status", pSet->label);
     (*pSet->fp)(data, pMonitor->dataLen, pMonitor->ip, 0, pSet->shandle, NULL, NULL);
     taosTmrReset(taosProcessMonitorTimer, 200, pMonitor, pSet->tmrCtrl, &pMonitor->pTimer);
-  }
-
-  if (pMonitor->pSet == NULL) {
+  } else {
     taosTmrStopA(&pMonitor->pTimer);
     free(pMonitor);
   }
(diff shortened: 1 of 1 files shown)


jtao1735:
Thanks for reading through the code. Although issues are not critical, but they can cause potential problems. 

For this small module, we plan to implement this feature: keep the TCP connection for a while after the data transfer is finished, so the TCP connection can be re-used for another new data transfer. It's like keep-alive feature in HTTP implementation, it can improve the system efficiency. 

Do you have interest and spare time to implement it? Thanks.

localvar:
I'd like to help on that and will give a try. however, I don't think I have enough understanding of the code currently and spare time is another issue, so please take my work as nice to have.

jtao1735:
Thanks to make the code more robust. May you leave your wechat or other IM ID? I would like to chat with you more on the TCP keep-alive feature

localvar:
sure. in fact, we exchanged wechat offline before, i will ping you later today.

PR #499, opened 2019-09-11
fix two memory leaks in meter creation

localvar:
a malicious client could bypass schema validation
to make the first column not a timestamp and trigger
the leak in `vnodeProcessCreateMeterMsg`.


localvar:
seems the fix not correct :(, will update later.

PR #600, opened 2019-10-09
Add support for continuous query in Lua connector.

robotspace:


robotspace:
上游代码分支结构按照git flow进行了修改，重新提交pull request 到develop分支。

Issue #642, opened 2019-10-24
druid pool support


jiaoyilun:
![image](https://user-images.githubusercontent.com/1704805/67448976-3e513400-f64b-11e9-84fc-565833409893.png)

PR #646, opened 2019-10-24
fix issue #642 to add  druid pool support

xieyinglin:
add  the druid pool support


xieyinglin:
edit the jdbc pom.xml for compile and edit the cmakeLists.txt  for copy different version jar

PR #696, opened 2019-11-11
[TBASE-1047] 

liu0x54:
Go connector prints all the SQL. This will harm the performance and consume unnecessary hard disk space. Modified to print only the SQLs which return error.


plum-lihui:
accept

PR #743, opened 2019-11-18
fix #742

### LINKED ISSUE #742: number of cache blocks may be messed up after importing a historical record into file
**Bug Description**
imported rows is 0 although there are some valid rows to be imported
key maybe messed up too, it causes the system to generate tons of files

**To Reproduce**
1: a cache block A for a table is not full, then commit, this cache block will be saved to disk. Firstkey in this cache block is F, last key is L, there are total T records. Since not full, this cache block wont be free 
2: after commit, insert some records, a new cache block B is allocated. Suppose R records are inserted into cache block A, the rest is in cache block B. 
3: import a record with key between F and L, R records will be moved to the head of cache block A, and there are T empty rows. 
4: import a record into cache block A, records in cache block B will be moved to block A.
5: Block B will have 0 points, but number of blocks is still 2. 

This will cause serious issue, since number of blocks is messed up.

jtao1735:


PEDM529:
https://github.com/users/jtao1735/achievements/pull-shark

PR #830, opened 2019-12-03
fix compile for raspbian on RaspberryPI 4B atomic not found error

Allenscript:
fix compile for raspbian on RaspberryPI 4B atomic not found error


guanshengliang:
@plum-lihui  请看一下这个编译选项

PR #1310, opened 2020-03-04
Fix timezone build issue on windows

sangshuduo:
src/util/src/ttime.c build failed on Windows due to timezone can't be found.


sangshuduo:
Only happen with msvc compiler but works with mingw. I guess what mingw provides is timezone and what msvc provides is _timezone makes different behavior.

PR #1323, opened 2020-03-08
Refact/rpc

jtao1735:


jtao1735:
change the message input/out parameters in RPC API  to SRpcMsg

PR #1355, opened 2020-03-12
remove unused comparison with 0 as size >=0 always

sangshuduo:
vi +220 src/system/detail/src/vnodeCommit.c

// restore from .olog file and commit to file
--
219 | size = vnodeRestoreDataFromLog(vnode, pVnode->logOFn, &firstV);
220 | if (size < 0) return -1;
  | Comparison is always false because size >= 0.
221 | if (size > 0) {
222 | if (pVnode->commitInProcess == 0) vnodeCommitToFile(pVnode);
↓↑ | 223-225
226 | // restore from .log file to cache
227 | size = vnodeRestoreDataFromLog(vnode, pVnode->logFn, &firstV);
228 | if (size < 0) return -1;
  | Comparison is always false because size >= 0.
229 |  
230 | if (pVnode->cfg.commitLog == 0) return 0;

### DIFF
--- a/src/system/detail/src/vnodeCommit.c
+++ b/src/system/detail/src/vnodeCommit.c
@@ -217,7 +217,6 @@ int vnodeInitCommit(int vnode) {
 
   // restore from .olog file and commit to file
   size = vnodeRestoreDataFromLog(vnode, pVnode->logOFn, &firstV);
-  if (size < 0) return -1;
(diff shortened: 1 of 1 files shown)


plum-lihui:
需要有返回-1的场景，所以还是修改一下函数的返回值类型。

PR #1399, opened 2020-03-22
fix the deadlock for double rpcCloseConn

jtao1735:


jtao1735:
fix the deadlock bug for rpcCloseConn

PR #1445, opened 2020-03-28
fix pMemBuffer memory leak and function name typo - /destory/destroy

sangshuduo:


plum-lihui:
到648行的时候，*pMemBuffer里面又malloc的内存，还不能简单的就释放*pMemBuffer？

sangshuduo:
Added additional memory free to avoid leak. Please review again.

PR #1658, opened 2020-04-19
Now able to run Python test against local server, without sudo/root privileges

stevel032:
No root/sudo needed now to run Python tests. Can you review and merge?


stevel032:
Thank you guys!!

PR #1784, opened 2020-05-03
[Trivial] Fix initialization code of TSDB_MOD_MGMT module.

haoyifan:
* Description
This patch fixes enablement of TSDB_MOD_MGMT module, which was
incorrectly assign to 'name' variable.

* Testing
Compile the code. Feel free to suggest more testing that needs to
be done. Given that 'tsModule' is zero initialized, I don't think
there's any functional difference for this patch.


haoyifan:
I didn't create an ISSUE to keep track of this change given it is pretty straightforward bug fix. Feel free to let me know if I need to associate this with an ISSUE.

PR #1788, opened 2020-05-04
[Vnode] Fix memory leak in vnodeProcessSubmitMsg

haoyifan:
Function call flow if msg type is TSDB_MSG_TYPE_SUBMIT:
dnodeProcessWriteQueue -> vnodeProcessWrite -> vnodeProcessSubmitMsg

vnodeProcessSubmitMsg() further calls rpcMallocCont() to allocate
memory for response message, but the allocation is never freed across
its life cycle, which causes memory leak.

Also, I don't see the response message used anywhere in the call stack,
and this is the same for other functions stored inside vnodeProcessWriteMsgFp.
Therefore this patch deliberately removes the allocation as well as the
message parameter altogether to eliminate the memory leak.

* Testing
Please suggest any test I can run to verify this problem. I guess
there's no stress test for message type of TSDB_MSG_TYPE_SUBMIT and
this problem slips your mind :)

### DIFF
--- a/src/vnode/src/vnodeWrite.c
+++ b/src/vnode/src/vnodeWrite.c
@@ -28,12 +28,12 @@
 #include "vnodeLog.h"
 #include "tcq.h"
 
-static int32_t (*vnodeProcessWriteMsgFp[TSDB_MSG_TYPE_MAX])(SVnodeObj *, void *, SRspRet *);
-static int32_t  vnodeProcessSubmitMsg(SVnodeObj *pVnode, void *pMsg, SRspRet *);
-static int32_t  vnodeProcessCreateTableMsg(SVnodeObj *pVnode, void *pMsg, SRspRet *);
-static int32_t  vnodeProcessDropTableMsg(SVnodeObj *pVnode, void *pMsg, SRspRet *);
-static int32_t  vnodeProcessAlterTableMsg(SVnodeObj *pVnode, void *pMsg, SRspRet *);
-static int32_t  vnodeProcessDropStableMsg(SVnodeObj *pVnode, void *pMsg, SRspRet *);
+static int32_t (*vnodeProcessWriteMsgFp[TSDB_MSG_TYPE_MAX])(SVnodeObj *, void *);
+static int32_t  vnodeProcessSubmitMsg(SVnodeObj *pVnode, void *pMsg);
+static int32_t  vnodeProcessCreateTableMsg(SVnodeObj *pVnode, void *pMsg);
+static int32_t  vnodeProcessDropTableMsg(SVnodeObj *pVnode, void *pMsg);
+static int32_t  vnodeProcessAlterTableMsg(SVnodeObj *pVnode, void *pMsg);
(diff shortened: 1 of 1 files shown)


src/vnode/src/vnodeWrite.c
```diff
@@ -77,36 +77,25 @@ int32_t vnodeProcessWrite(void *param1, int qtype, void *param2, void *item) {
...
 
-  pRet->len = sizeof(SShellSubmitRspMsg);
-  pRet->rsp = rpcMallocCont(pRet->len);
-  SShellSubmitRspMsg *pRsp = pRet->rsp;
-
-  pRsp->code              = 0;
-  pRsp->numOfRows         = htonl(1);
-  pRsp->affectedRows      = htonl(1);
-  pRsp->numOfFailedBlocks = 0;
-
```
jtao1735, line 105: these a few lines shall no t be removed, since TDengine client needs to know affected rows and other information besides the code (success or an error code). That's why we need SRspRet parameter. Create/Alter/Drop Table/DB may return some info to client even they don't return anything in current version.

haoyifan: Looks like SRspRet is overwritten immediately in dnodeProcessWriteQueue() instead of returning back to the RPC module. Am I missing something here?

```
for (int32_t i = 0; i < numOfMsgs; ++i) {
      pWrite = NULL;
      taosGetQitem(pWorker->qall, &type, &item);
      if (type == TAOS_QTYPE_RPC) {
        pWrite = (SWriteMsg *)item;
        pHead = (SWalHead *)(pWrite->pCont - sizeof(SWalHead));
        pHead->msgType = pWrite->rpcMsg.msgType;
        pHead->version = 0;
        pHead->len = pWrite->contLen;
      } else {
        pHead = (SWalHead *)item;
      }

      int32_t code = vnodeProcessWrite(pVnode, type, pHead, item); <-- item gets overwritten within this loop
      if (pWrite) pWrite->rpcMsg.code = code;
    }

    walFsync(vnodeGetWal(pVnode));

    // browse all items, and process them one by one
    taosResetQitems(pWorker->qall);
    for (int32_t i = 0; i < numOfMsgs; ++i) {
      taosGetQitem(pWorker->qall, &type, &item); <-- item also gets overwritten here
      if (type == TAOS_QTYPE_RPC) {
        pWrite = (SWriteMsg *)item;
        dnodeSendRpcWriteRsp(pVnode, item, pWrite->rpcMsg.code); 
      } else {
        taosFreeQitem(item);
        vnodeRelease(pVnode);
      }
    }
```

jtao1735: it wont be overwritten, since a new pWrite will be retrieved from taosGetQitem in each loop.
SRspRet is part of SWriteMsg, so the response can be passed out.

haoyifan: Ah I see. I missed that part of the code. Thanks for the explanation :)

haoyifan:
Not a fix.

PR #1798, opened 2020-05-05
[Dnode] Table driven method to init and cleanup dnode components.

haoyifan:
This patch refactors the initialization and cleanup path of dnode
with a table driven method. This fixes the following issues:

1. Before the patch, if dnodeInitRead() fails, the cleanup path
also runs dnodeCleanupWrite(), which will free wWorkerPool.writeWorker that's
never allocated. (The code before this patch will not crash though, because
wWorkerPool is zero-initialized global variable and therefore the
accidental free will be a nop).
2. In general the order of calling cleanup function should be reverse
to the order of calling init function, but this is not the case prior
to this patch (see dnodeCleanupMnode() and dnodeCleanupMgmt()).

* Bonus fix
This patch also fixes a missing free for readPool.readWorker.
This patch also removes a redundant taosCloseLog() in cleanup path.

* Testing
I plan to run the test script ./test-all.sh, but was not able to do so.
Is there a reference somewhere I can look up?

### DIFF
--- a/src/dnode/src/dnodeMain.c
+++ b/src/dnode/src/dnodeMain.c
@@ -41,16 +41,16 @@ static void dnodeSetRunStatus(SDnodeRunStatus status);
 static void signal_handler(int32_t signum, siginfo_t *sigInfo, void *context);
 static void dnodeCheckDataDirOpenned(char *dir);
 static SDnodeRunStatus tsDnodeRunStatus = TSDB_DNODE_RUN_STATUS_STOPPED;
-static int32_t dnodeInitSteps();
-static void dnodeCleanupSteps(int32_t stepId);
+static int32_t dnodeInitComponents();
+static void dnodeCleanupComponents(int32_t stepId);
 
 typedef struct {
   const char *const name;
   int               (*init)();
   void              (*cleanup)();
-} DnodeStep;
+} SDnodeComponent;
 
-static const DnodeStep DnodeSteps[] = {
+static const SDnodeComponent SDnodeComponents[] = {
   {"storage",       dnodeInitStorage,       dnodeCleanupStorage},
   {"read",          dnodeInitRead,          dnodeCleanupRead},
   {"write",         dnodeInitWrite,         dnodeCleanupWrite},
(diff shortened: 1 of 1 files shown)


src/dnode/src/dnodeMain.c
```diff
@@ -41,6 +41,25 @@ static void dnodeSetRunStatus(SDnodeRunStatus status);
...
 static void dnodeCheckDataDirOpenned(char *dir);
 static SDnodeRunStatus tsDnodeRunStatus = TSDB_DNODE_RUN_STATUS_STOPPED;
+static int32_t dnodeInitSteps();
+static void dnodeCleanupSteps(int32_t stepId);
+
+typedef struct {
+  const char *const name;
+  int               (*init)();
+  void              (*cleanup)();
+} DnodeStep;
```
jtao1735: For TDengine convention,  structure name starts with Upper case 'S'. Also, may you change the name to SDnodeModule.  It makes more readable.

src/dnode/src/dnodeMain.c
```diff
@@ -135,6 +152,24 @@ static void signal_handler(int32_t signum, siginfo_t *sigInfo, void *context) {
   exit(EXIT_SUCCESS);
 }
 
+static void dnodeCleanupSteps(int32_t stepId) {
```
jtao1735: rename "dnodeCleanupSteps" to "dnodeCleanupComponents". In addition, remove the input parameter "stepId", it shall be similar to other init/cleanup functions

src/dnode/src/dnodeMain.c
```diff
@@ -135,6 +152,24 @@ static void signal_handler(int32_t signum, siginfo_t *sigInfo, void *context) {
...
+  for (int32_t i = stepId; i >= 0; i--) {
+    DnodeSteps[i].cleanup();
+  }
+}
+
+static int32_t dnodeInitSteps() {
```
jtao1735: renname to dnodeInitComponents

jtao1735:
For the testing, we will publish a set of scripts soon.

src/dnode/src/dnodeMain.c
haoyifan: The function introduces a *stepId* because while it cleanup during failure, you don't want it to run cleanup() function where its corresponding init() never runs. For example if the readInit fails, you only want the unwinding starts from readCleanup and backwards.

haoyifan:
> For the testing, we will publish a set of scripts soon.

I was going to ask, how do you run test-all.py? I tried to run it but it fails with import of 'taos' module, even after I compiled and installed the software.

src/dnode/src/dnodeMain.c
haoyifan: I name the struct as SDnodeComponent to make it consistent with the init and cleanup function, and also to eliminate confusion since the code already has 'modules'.

src/dnode/src/dnodeMain.c
jtao1735: Yeah, a good way to handle this scenario

localvar:
for table driven init/uninit, debugging also need to be considered: generally,it makes debugging difficult because one can not find out which function is being called directly from the code. so the best scenario for table driven is plugins,because functions to be called are not known at the time the caller is written. but if we know all/most of the callee function, call them one by one would be better, though looks a little ugly.

haoyifan:
> for table driven init/uninit, debugging also need to be considered: generally,it makes debugging difficult because one can not find out which function is being called directly from the code. the best scenario for table driven is plugins, where functions to be called are unknown at the time the caller is written. but if we know all/most of the callee functions, call them one by one would be better, though looks a little stupid, but it is an example of the KISS principle.
> 
> I've written a blog post(in Chinese) about this before,please refer https://blog.localvar.cn/code-maintainability-2 for more information.

Thanks for linking your blog :)

My initial goal to make it table driven is that let's say we have initA()/cleaupA() and initB()/cleanupB(). If initA() fails, we only want to call cleanupA() but not cleanupB(), because cleanupB might try to cleanup something that's not yet initialized in initB(). If it is guaranteed that calling cleanup function is ok even if its corresponding init() is never called, then we should have no problem with the existing approach. Table driven method simply makes rolling back in a halfway failure simpler.

PR #1809, opened 2020-05-06
[vnode] Make vnode versioning more robust.

haoyifan:
This patch changes version read/write path such that it does not
allocate heap memory while dealing with versioning. Instead, because
version data is short, it uses stack allcocated array to hold the versioning
data. This eliminate one potential failure in read/write version, which is
used to guarantee data consistency.

* Bonus fix
The patch also fixes a bunch of failure handling in the path of vnode
write.


jtao1735:
We plan to have some structure changes in the coming two weeks. Please hold a while.

jtao1735:
Please check the latest code in develop branch. vnodeOpen and vnodeCleanUp are refactored, so all the issues you found are fixed. Thanks.

PR #2015, opened 2020-05-23
Implemented subscription json data from Mqtt broker! 

maikebing:
- [x] Add plug-in basics

- [x] Add Mqtt client 

- [x] Add configuration item for  Mqtt broker's  server address 

- [x] Add configuration item for  Mqtt broker's   user, password 
 
- [x] Convert json to sql

- [x] Subscribe  json data from  Mqtt broker


guanshengliang:
According to the previous coding standards:
1. The functions included in mqttSystem.h are called for other modules, so the function declarations such as `split` `onDisconnectFailure `, should not be in mqttSystem.h, but in other internal header files, ex: `mqttInt.h`.
2. Because there is no namespace in C language, the functions of mqtt module should be prefixed with mqtt, such as `connlost`, `mqtt_query_insert_callback`, modified to `mqttConnLost`, `mqttQueryInsertCallback`
3. If the function is only used in one single C file, you can use static to declare the function

guanshengliang:
In addition, because the 2.0 version will be released soon, the develop branch code is submitted frequently, you can merge before submit to avoid conflicts

stevel032:
Dear Maike... we saw quite a bit of your submission has code under the Eclipse Public License, which is a so called "copy left license", requiring us to disclose all our source code. Plus, it requires all the users of our own software to do this safe. Generally we stay away from such licenses, and as such we cannot accept you submission as is. We hope you understand, and would appreciate that you continue to make contributions!

maikebing:
@stevel032  有规避方法吗？比如引用git子模块的方式？

stevel032:
Thank you for the suggestion. I wish there were easy ways, but there are not.

The copy-left series of licensing scheme has been around for 20+ years at least, and it's been a well established practice in the industry and there's been no consensus on how to "get around" it. Further more, it reflects the desire of the author of the software, and such desire should be respected, not the least of which among us here at TaosData.

If we really want to figure out an alternative though, an industry practice closest to "getting around" such restrictions is to contain the copy-left code in stand-alone binaries, in the form of a "plug in" that can be dynamically utilized by the system at run time. Then the customer/user of the system needs to obtain the plug-in directly. It's a convoluted process, and we are not anywhere close to being in that situation.

So long story short... we really can't do this right now.

maikebing:
我找一下 其他许可的 稳定的mqtt客户端实现 吧

maikebing:
MIT协议的没问题吧？

stevel032:
Yes, MIT licenses are much more amenable to us. Let us review this soon.

Thank you so much for your continued attention and contribution!

Issue #1989, opened 2020-05-21
strncpy broken


freemine:
in src file: src/mnode/src/mgmtTable.c
1307         strncpy(pVgroupInfo->vgroups[vgSize].ipAddr[vn].fqdn, pDnode->dnodeFqdn, tListLen(pDnode->dnodeFqdn));

size of dnodeFqdn is (TSDB_FQDN_LEN+1) while
size of fqdn is (TSDB_FQDN_LEN)

guanshengliang:
Thanks for the suggestion, we will use macro tstrncpy to deal with similar problems, such as 
tstrncpy(pVgroupInfo->vgroups[vgSize].epAddr[vn].fqdn, pDnode->dnodeFqdn, TSDB_FQDN_LEN);

guanshengliang:
This issue has been resolved, so we close it

PR #2020, opened 2020-05-24
fix some of strncpy calls, specifying n as length of dest buf

freemine:
#1989 

strncpy with `moving` pointer is not touched in this PR, such as those in tVariantXXX functions.

### DIFF
--- a/src/plugins/http/src/httpAuth.c
+++ b/src/plugins/http/src/httpAuth.c
@@ -91,7 +91,7 @@ bool httpParseTaosdAuthToken(HttpContext *pContext, char *token, int len) {
 
     // for the purpose of httpTrace, but hereby, buffer overflow!!
     // to handle this problem, we use tmp to save and restore
-    const char tmp = pContext->pass[sizeof(pContext->pass)];
+    const char tmp = *(pContext->pass+sizeof(pContext->pass));
     *(pContext->pass+sizeof(pContext->pass)) = 0;
     httpTrace("context:%p, fd:%d, ip:%s, taosd token:%s parsed success, user:%s", pContext, pContext->fd,
               pContext->ipstr, token, pContext->user);


localvar:
for the use of `tListLen`, though the result is correct, this macro is to calculate element count of an array instead of byte count while what `strncpy` required is byte count.

please replace it with `sizeof`, thanks.

freemine:
@localvar 
yes, better use sizeof instead of tListLen. I chose to use tListLen because there have already tListLen used in previouse strncpy calls. Will revert to sizeof in next trial submit.

for the previous failed submits, it's been found out that some of strncpy calls are really for binary data / not-null terminated ones.

will try to pass the local tests before submitting.

src/util/inc/tutil.h
```diff
@@ -110,6 +110,23 @@ extern "C" {
...
+
+#define STRNCPY(dst, src, n)       \
+  do {                             \
+    assert((n) <= sizeof(dst));    \
+    int nn = n;                    \
+    if (nn >= sizeof(dst)) {       \
+      nn = sizeof(dst) - 1;        \
+    }                              \
+    strncpy((dst), (src), nn);     \
+    (dst)[nn] = 0;                 \
```
localvar, line 134: if `n` is less than `sizeof(dst)`, this line modifies the `n + 1` byte, which is not expected.

src/util/src/ttimer.c
```diff
@@ -552,7 +552,7 @@ void* taosTmrInit(int maxNumOfTmrs, int resolution, int longest, const char* lab
     return NULL;
   }
 
-  strncpy(ctrl->label, label, sizeof(ctrl->label));
+  STRNCPY(ctrl->label, label, sizeof(ctrl->label));
   ctrl->label[sizeof(ctrl->label) - 1] = 0;
```
localvar: this line could be removed after the change.

src/util/inc/tutil.h
```diff
@@ -110,6 +110,23 @@ extern "C" {
...
 
+#define MEMNCPY(dst, src, n)       \
+  do {                             \
+    assert((n) <= sizeof(dst));    \
+    strncpy((dst), (src), n);      \
+  } while (0)
+
+#define STRNCPY(dst, src, n)       \
+  do {                             \
+    assert((n) <= sizeof(dst));    \
```
localvar, line 128: it is not correct to use `sizeof(dst)` here if `dst` is a pointer instead of an array.

src/util/inc/tutil.h
freemine: please be kindly noted, a MEMNCPY macro is also added (just lines before STRNCPY definition) which does not write null-terminator, and you can use MEMNCPY to handle the case you mentioned.

src/util/src/ttimer.c
freemine: ok

src/util/inc/tutil.h
freemine: yes, I know what you mean. Because it's C, I have no decent way to distinguish whether it's pointer or array. what else would you suggest, if you happen to know?

thx!

localvar: pls revise `STRNCPY` as below and remove `MEMNCPY`:
```c
#define STRNCPY(dst, src, n) \
do { \
   strncpy((dst), (src), (n)); \
   (dst)[(n) - 1] = 0; \
} while (0);
```

the idea behind is: we should try to avoid behavior changes when redefine a standard API.

because what we need here is a `strncpy` like function but could guarantee destination string is null terminated, the behavior change should be limit to "null terminated", otherwise, it will be very possible for the new function/macro to be misused, as developers think it is just a `strncpy`。

 `MEMNCPY` is similiar, developers will think it is a `memcpy`, but it is a `strncpy`.

freemine: yes, i know the idea behind. but i still have some uncertainties as following:
just take `src/client/src/tscSql.c` as example:
line 89:  `strncpy(pObj->user, user, TSDB_USER_LEN);`
whereby `pObj->user` is defined in `src/client/inc/tsclient.h` as 
284 typedef struct STscObj {
...
289   char               user[TSDB_USER_LEN];
...
this seems also a bug, will user is TSDB_USER_LEN long or TSDB_USER_LEN-1 long?

and then, we take another example in `src/kit/taosdump/taosdump.c`
line 391:  `strncpy(pTableRecordInfo->tableRecord.name, (char*)row[TSDB_SHOW_TABLES_NAME_INDEX], fields[TSDB_SHOW_TABLES_NAME_INDEX].bytes);`
again: is ...bytes including null-terminator or not?

having these two concerns in mind, I chose those STRNCPY/MEMNCPY as a preliminary solution, with assert to check (in debug mode), and never overwrite dest buffer.

in order to make the coding phylosophy consistent, your idea shall be followed as suggested. but i am afraid of being less clear about the program logic under the hood,  submission as suggestion might fail the test (for those boundary test cases).

even if i chose to change user[TSDB_USER_LEN]; to user[TSDB_USER_LEN+1];
that might potentially break somewhere else? i have no idea.

thus, MEMNCPY/STRNCPY, as defined this way temporarily, it'll introduce code-break as less as possible.

as a compromise, shall we submit this way first, and then check boundary-failure test later, and make a final decision?

sorry for this lengthy response.

freemine: btw, look at those failed submissions under this pr, that tells the story that how less familiar that i am with the logic underneath till now. :)

localvar: I totally understand your feeling, because it is also my feeling.

generally, when define the max length of a string, the terminator should be included. but in current code base, sometimes the terminator is included while sometimes not, and this is even happen in one variable!

however, this is not a reason for us to break the behavior of standard functions, because it will make things worse.

if some test cases do fail, we just fix them, from a long term view, this way makes the code better.

freemine: ok

freemine:
:(
all failed as expected.

localvar:
hi @freemine , I'm very sorry but I have to close this PR, thanks very much for your efforts.

we decide to make a universal fix for string length related issues, and there will be too many conflicts if we accept this PR.

our decision cannot be made without your push in this PR, really appreciate. hope you will continue support us, thanks.

PR #2173, opened 2020-06-07
early allocation of epoll notification fd

freemine:
early allocate epoll fd to minimize the probability of runtime failure during tcp connection shutdown

### DIFF
--- a/src/rpc/src/rpcTcp.c
+++ b/src/rpc/src/rpcTcp.c
@@ -132,6 +132,7 @@ void *taosInitTcpServer(uint32_t ip, uint16_t port, char *label, int numOfThread
       pThreadObj->fd_event     = eventfd(1, 0);
       if (pThreadObj->fd_event < 0) {
         tError("%s failed to create fd_event for epoll event", label);
+        terrno = TAOS_SYSTEM_ERROR(errno);
         code = -1;
         break;
       }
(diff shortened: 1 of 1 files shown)


freemine:
I refetched repo again, and found there's no conflict.
i also noticed PR #2172. I can wait until #2172 is merged then.

guanshengliang:
@jtao1735  just modified this file
now there is some conflict
please check and initiate a new PR, then let @jtao1735  review

freemine:
thx.

the latest code for cleanup in case of early init failure may suffer intialized mutex un-destroyed (if numThreads are modified) or destroying an uninitialized mutex(if otherwise). but making code simple and clear is a good suggestion. i'll make some modification upon this.

as for kqueue/epoll, i searched the code and found src/plugins/http/src/httpServer.c is also relying on epoll. are we considering to port that code too?

freemine:
as for kqueue/epoll issue:
seems like currently, tdengine is servicing one connection per thread. thus, kqueue/epoll does not bring any benefit.

if it's decided to switch to multiplex mode, why not relay on libevent/libev/libuv?

jtao1735:
> as for kqueue/epoll issue:
> seems like currently, tdengine is servicing one connection per thread. thus, kqueue/epoll does not bring any benefit.
> 
> if it's decided to switch to multiplex mode, why not relay on libevent/libev/libuv?

We planned to implement kqueue at the beginning of the project, but never got free time to implement it. Also, we planed to adopt DPDK for more efficient solution, you can do some research on it.

PR #2196, opened 2020-06-09
Patch/string length

localvar:


freemine:
IMHO, i really suggest xxx_strncpy(darray, src, sizeof(dst)) rather than xxx_strncpy(darray, src, TSDB_XXX_LEN).

it's easy for the reviewer/maintainer know it's safe call here. leaving TSDB_XXX_LEN dangling all around source codes, it's really distraction and problem-prone

back to header file declaration, we know that some would be defined as buf[TSDB_XXX_LEN] while the other buf[TSDB_XXX_LEN+1]. normally speaking, the former is usually reflecting file storage layout or network payload, while the latter means it's memory usage only, very much likely to be used with printf sort of calls.

thus:
for fs => mem: use xxx_strncpy(darray, src, sizeof(dst)); darray[sizeof(dst)-1] = 0;
for mem => fs: use xxx_strncpy(darray, src, sizeof(dst)); leaving dst not null-terminating

for codes where in patterns like:
writing(big part) {
  writing subpart1
  writing subpart2
  ...
}
they could be refactor as such:
writing(*p, bytes) {
  n = xnprintf(p, bytes, ...)
  p += n; bytes -= n;
  ....
}

src/client/src/tscParseInsert.c
```diff
@@ -989,7 +989,7 @@ static int32_t tscCheckIfCreateTable(char **sqlstr, SSqlObj *pSql) {
 
 int validateTableName(char *tblName, int len) {
   char buf[TSDB_TABLE_ID_LEN] = {0};
-  strncpy(buf, tblName, len);
+  tstrncpy(buf, tblName, sizeof(buf));
```
freemine, line 992: shall we redefine buf[TSDB_TABLE_ID_LEN] to buf[TSDB_TABLE_ID_LEN+1] ?
otherwise, it might conflict with the doc which says table id len is as TSDB_TABLE_ID_LEN.

localvar: this patch is still work in progress,I create this PR to submit part of the changes to avoid conflicts with other changes.
TSDB_TABLE_ID_LEN related issues will be fixed later.

and thanks for you suggestion, most of the new code are using `sizeof` instead of `TSDB_XXX_LEN` now.

freemine:
latest commit's broken, cae80f403ef02e596c4dcea7dcc46182c3cf4a58
eg.: 
src/client/src/tscLocal.c
```
   char* vx = calloc(1, pExpr->resBytes);
-  STR_WITH_MAXSIZE_TO_VARSTR(vx, pSql->pTscObj->user, TSDB_USER_LEN);
+  STR_WITH_MAXSIZE_TO_VARSTR(vx, pSql->pTscObj->user, sizeof(pSql->pTscObj->user));
```
compiler complains about using sizeof(src) rather than sizeof(dst).

this makes me recall my previous version of macros:

```
// note: copy at most n-bytes of src to dst if n is less than sizeof(dst)
//       otherwise, copy sizeof(dst) of src to dst
//       null-terminated?: not defined
#define MEMNCPY(dst, src, n)       \
  do {                             \
    assert((n) <= sizeof(dst));    \
    strncpy((dst), (src), n);      \
  } while (0)

// note: copy at most n-bytes of src to dst if n is less than sizeof(dst)
//       otherwise, copy sizeof(dst)-1 of src to dst
//       null-terminated in both case
// note1: in either case, bytes beyond n in src, is never accessed at run time!
#define STRNCPY(dst, src, n)       \
  do {                             \
    assert((n) <= sizeof(dst));    \
    int nn = n;                    \
    if (nn >= sizeof(dst)) {       \
      nn = sizeof(dst) - 1;        \
    }                              \
    strncpy((dst), (src), nn);     \
    (dst)[nn] = 0;                 \
  } while (0)
```

these definitions take both src and dst in concern, with STRNCPY having dst null-terminated while MEMNCPY none.

At that moment, i did not notice VARCHAR struct, I took it for granted that tdengine would have it's own function-based VARCHAR implementations.

btw, i would not suggest STR_WITH_MAXSIZE_TO_VARSTR macro is defined in this way, or, varchar variable initialization/modification is implemented in this way.

just share my humble opinion.

freemine:
i am confused by the recent commits concerning about the TSDB_XXX_LEN. let's take
TSDB_COL_NAME_LEN as example. it's  formally defined as 64, but now it's 65.
i have not checked the code yet, but i believe this modification will change the datafile layout, which will result in db version compatibility problem, i guess.
if relevent code has been modified accordingly, compatibility problem will not occur. but how would that 65 mean to tdengine's developer/users? how would tdengine denotes column name max length in your official document to end users, who just use sql and have no idea about c programming? how would any specific person tell what's the difference between such lines copied from that src/inc/taosdef.h:
```
#define TSDB_ACCT_LEN             11
#define TSDB_NODE_NAME_LEN        64
#define TSDB_TABLE_NAME_LEN       193     // it is a null-terminated string
#define TSDB_DB_NAME_LEN          33
#define TSDB_TABLE_ID_LEN         (TSDB_ACCT_LEN + TSDB_DB_NAME_LEN + TSDB_TABLE_NAME_LEN)
#define TSDB_COL_NAME_LEN         65
```

can any one tell what's the length of TSDB_TABLE_ID_LEN? any logical explanation?
TSDB_TABLE_ID_LEN has three null-terminators in total? or just wish c code does not core-dump?

freemine:
it seems to me, that db/infrasture designer holds his responsibility to define lengths of those fields in their logic way. and then, it's our programer's job to fulfill this in a correct and decent way.

but now i see too many are involved in this critical issue which bring things worse than before.

pls forgive my straight forward. :(

PR #2227, opened 2020-06-10
log print error

guanshengliang:


src/dnode/src/dnodeMgmt.c
```diff
@@ -428,7 +428,7 @@ static bool dnodeCheckMnodeInfos(SDMMnodeInfos *pMnodes) {
   for (int32_t i = 0; i < pMnodes->nodeNum; ++i) {
     SDMMnodeInfo *pMnodeInfo = &pMnodes->nodeInfos[i];
     if (pMnodeInfo->nodeId <= 0 || strlen(pMnodeInfo->nodeEp) <= 5) {
-      dError("invalid mnode info:%d, nodeId:%d nodeEp:%s", pMnodeInfo->nodeId, pMnodeInfo->nodeEp);
+      dError("invalid mnode info:%d, nodeId:%d nodeEp:%s", i, pMnodeInfo->nodeId, pMnodeInfo->nodeEp);
```
freemine, line 431: to solve such kind of probem, you might think of using macro to let compiler check in early time.
eg:
void xxx_log(const char *buf);

#define LOG(fmt, ...) do {
   char buf[4096];
   snprintf(buf, fmt, ##__VA__ARGS__);
   xxx_log(buf);
} while (0)
// if you need __FUNC__,__FILE__,__LINE__, they could be all placed in macro.

in this way, if there is any type-mismatch, count of placeholders-mismatch, you'll get warning message from compiler, rather than leaving it at runtime, which at the worst case is a security pitfall.

PR #2233, opened 2020-06-11
[td-582] fix crash.

hjxilinx:


freemine:
IMHO, I really suggest, refactory header files first, be careful about the difference between buf[TSDB_XXX_LEN] and buf[TSDB_XXX_LEN+1].

then go back to source files, replace arbitrary len/size/strlen with sizeof, choose tstrncpy/memcpy carefully accordingly.

if there happen to be some local variable defined as buf[TSDB_.....], please make it consistent with the header file, if that varaible is supposed to work as a string, use buf[TSDB__xxx_LEN+1]
mode.

otherwise, refactory would be frustrating and problem-prone.

wish u luck

hjxilinx:
Thanks for your suggestions. 
This work is undergoing. We will remove any existed local variables or attributions of struct in header files or .c files in the format like TSDB_XXX_LEN+1.

PR #2330, opened 2020-06-17
make stop taosd by kill always done with continue check

sangshuduo:
[TD-670]


tests/script/general/parser/limit1_tblocks100.sim
```diff
@@ -61,11 +61,10 @@ run general/parser/limit1_stb.sim
 
 print ================== restart server to commit data into disk
 system sh/exec.sh -n dnode1 -s stop -x SIGINT
-sleep 5000
 system sh/exec.sh -n dnode1 -s start
```
guanshengliang, line 64: Are you sure you don’t want to sleep at this line

sangshuduo: no need sleep to wait since the sh/exec.sh will done to make sure taosd be killed.

PR #3033, opened 2020-08-12
TD-1057 set crash in windows

guanshengliang:


localvar:
`MiniDumpWithHandleData` does not dump all memory and result in a smaller dump file, need to change to `MiniDumpWithFullMemory` for full memory dump.

PR #3215, opened 2020-08-24
install.sh: create empty history file after installation

amoskong:
There is a error message in first use of taos, because the history
file doesn't exist.

This patch tried to create an empty history file after installation,
then users won't see the error message.

| [root@045a9307c53a ~]# taos
|
| Welcome to the TDengine shell from Linux, Client Version:2.0.1.1
| Copyright (c) 2017 by TAOS Data, Inc. All rights reserved.
|
| Failed to open file /root/.taos_history
| taos>

Signed-off-by: Amos Kong <kongjianjun@gmail.com>

### DIFF
--- a/packaging/tools/install.sh
+++ b/packaging/tools/install.sh
@@ -713,6 +713,7 @@ function install_TDengine() {
         echo
         echo -e "\033[44;32;1mTDengine client is installed successfully!${NC}"
     fi
+    touch ~/.taos_history
 
     rm -rf $(tar -tf taos.tar.gz)
 }


plum-lihui:
when  Taos initialization , judge that if no history file does not exist, it is better to remove the output information.

amoskong:
> when Taos initialization , judge that if no history file does not exist, it is better to remove the output information.

If the file doesn't exist, touch will create an empty file.  If the file exists the file content won't be changed, just update the timestamps of this file.

`touch` command won't output anything if the ~/.taos_history exists or not.

PR #3267, opened 2020-08-27
TD-857

yihaoDeng:
failure to filter tag column of binary type


guanshengliang (approved):

yihaoDeng:
if first tag column is binary,  query like  'select * from meters where first_tag like '%X'' failed to filter valid subtable, this PR fix this problem

PR #3296, opened 2020-08-30
Update query api for Lua connector.

robotspace:
Update query api for Lua connector.
Add entrance for Lua connector introduction.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ The TDengine community has also kindly built some of their own connectors! Follo
 
 - [Rust Connector](https://github.com/taosdata/TDengine/tree/master/tests/examples/rust)
 - [.Net Core Connector](https://github.com/maikebing/Maikebing.EntityFrameworkCore.Taos)
+- [Lua Connector](https://github.com/taosdata/TDengine/tree/develop/tests/examples/lua)
 
 # How to run the test cases and how to add a new test case?
   TDengine's test framework and all test cases are fully open source.


guanshengliang (approved):

sangshuduo:
@robotspace hi, thanks for your contribution to add lua connector for TDengine. Could you please introduce how we can verify/test this lua connector works as expected or not? Then we would add the test case to make sure the connector works well in the future even something changed in TDengine.

robotspace:
> @robotspace hi, thanks for your contribution to add lua connector for TDengine. Could you please introduce how we can verify/test this lua connector works as expected or not? Then we would add the test case to make sure the connector works well in the future even something changed in TDengine.

Thanks very much for your concern. The connector is implemented by Lua version5.3, so you may need to install Lua to execute the build.sh script and run test.lua. I have designed some cases such as creating table and inserting record, etc. I don't think this is the best way to verify connector. The key question is how to define pass/fail and get test result from tdengine's auto-test kit. It's quite a problem for me right now. I will try my best to find a solution to resolve it and send a blog to TAOS.

sangshuduo:
@robotspace good to know your blog is coming. My email is sdsang@taosdata.com. Thanks.

robotspace:
> @robotspace good to know your blog is coming. My email is [sdsang@taosdata.com](mailto:sdsang@taosdata.com). Thanks.

Got it.

PR #3539, opened 2020-09-16
[TD-1468]

plum-lihui:
[TD-1468]: Error print to open faile when firstly read .taos_history.


plum-lihui:
[TD-1468]:<fix> Error print to open faile when firstly read .taos_history.

guanshengliang (approved):

PR #3585, opened 2020-09-18
Added valgrind memory check to crash_gen tool

stevel032:
Used to report TD-1523, now able to display TAOS memory leak issues with Python script. The check-in also included a so called Valgrind error suppression file, generated using the valgrind tool itself, as the Python run time does not always clean up all the memory it uses. Without such suppression file, the TAOS related memory leaks will be hard to see among the numerous Python memory leaks.


stevel032:
Just added further enhancements:

- With the "-b N" option, the tool is able to operate against multiple databases concurrently.
- With the "--valgrind" option, the tool is able to run under the valgrind tool, and produce capture the output/error message in redirected files.

PR #3616, opened 2020-09-21
fix binary length

plum-lihui:
[TD-1503] <fix> fix bug for binary length in taos go driver.


Kittoa:
Thank you very much, I just ran into this problem

PR #2970, opened 2020-08-07
Refactor tcache

freemine:
as for issue: #2969

this PR is supposed to tackle this. please review and comment.

thx!

PR #3695, opened 2020-09-27
add a new hash implementation, both thread-safe and reentrant-available

freemine:
#2970 

**1. add ehash, a thread-safe/reentrant hash implement
2. reimplement hash.c with ehash, with hash api unchanged, neither api sig nor semantics
3. #2 is for back-compatibility, though ehash is believed a better one
4. with pthread_once/atexit, make taosCloseLog elegant in tearing-down
5. add edge-test-cases**

btw: add USE_EHASH and TRACE_HASH option in CMakeLists.txt.
USE_EHASH is by default ON, which means ehash would be used and original hash would be wrapped with ehash.
TRACE_HASH is by default OFF, enabling this would log HASH API on SCREEN, which might help you in debug

if you wanna fallback to original hash, you can:
```
rm -rf debug && cmake -DUSE_EHASH=OFF -B debug && cmake --build debug
```

### DIFF
--- a/packaging/docker/Dockerfile
+++ b/packaging/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM ubuntu:16
+FROM ubuntu:20.04
 
 WORKDIR /root
 
@@ -7,7 +7,7 @@ RUN echo $version
 COPY tdengine.tar.gz /root/
 RUN tar -zxf tdengine.tar.gz
 WORKDIR /root/TDengine-server-$version/
-RUN sh install.sh -e no
+RUN /bin/bash install.sh -e no
 
 
 ENV LD_LIBRARY_PATH="$LD_LIBRARY_PATH:/usr/lib"
--- a/packaging/docker/dockerbuild-aarch64.sh
+++ b/packaging/docker/dockerbuild-aarch64.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -x
+docker build --rm -f "Dockerfile" -t tdengine/tdengine-aarch64:$1 "." --build-arg version=$1
+docker login -u tdengine -p $2  #replace the docker registry username and password
+docker push tdengine/tdengine-aarch64:$1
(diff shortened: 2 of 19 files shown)


freemine:
there's no windows env by hand, thus `basename` failure could not be handled on my side.

freemine:
截屏2020-09-28 13 00 09
what does this mean？related to this PR or not?

sangshuduo:
@freemine could you please run "./crash\_gen.sh -a -p -t 4 -s 2000" in TDengine/test/pyteset directory? Your PR failed in CI with this test case.

sangshuduo:
@freemine could you please run "./crash\_gen.sh -a -p -t 4 -s 2000" in TDengine/test/pyteset directory? Your PR failed in CI with this test case.

freemine:
yes, noticed that and is working on it

freemine:
now, i tent to believe, assertion reported by CI is coming directly from the complex logic in tcache.c.

it's too late, i'll check it later, if no one happens to find the root cause. :)

freemine:
btw, with this handy valgrind-enabled crash_gen.py, some of 'unintialized data uses' were detected in tsdbRead.c

freemine:
@sangshuduo

src/util/inc/tcache.h
```diff
@@ -54,6 +54,7 @@ typedef struct SCacheDataNode {
   uint32_t           size;         // allocated size for current SCacheDataNode
   T_REF_DECLARE()
   char              *key;
+  unsigned int       data_freed:2;
```
hjxilinx, line 57: I am not quit clear why "data_freed" is needed? Could you please explain it a litttle bit.

freemine: it's been a while since last PR'd.

please be kindly noted:
```
137   if (pCacheObj->freeFp) {
138     DASSERT(pElem->pData->data_freed==0);
139     if (!pElem->pData->data_freed) {
140       pCacheObj->freeFp(pElem->pData->data);
141       pElem->pData->data_freed = 1;
142     }
143   }
```
DASSERT is there to help find `double-free` problems.

well, it exists because cached object could be freed within cacher-timer-thread or user-thread if we can say so, we
need a way to prohibit freeFp from being called more than once. from the critical point of view, `data-freed` is not as good as expected. 


this PR's main purpose was to re-implement a better hash. modifications in tcache.c, including data-freed, is not
a firm and solid one. and i am still whatsoever confused by the cache api semantics brought in tache.h. because tcache's been used heavenly in the project, i hesitated to modify tcache at the very beginning.

wish this helps.

PR #3741, opened 2020-09-30
TD-1632

jtao1735:
race condition for pConn


jtao1735:
remove a race condition

guanshengliang (approved):

guanshengliang (approved):

PR #3931, opened 2020-10-20
[TD-1471]<fix>: empty result was returned if a database was dropped during query

localvar:


stevel032:
Travis reports following problem from the crash_gen tool:

```
errno=0x201, msg: Invalid qhandle, SQL: select tbname from db_0.fs_table where tbname in ('reg_table_1')
```

Maybe we should look at the code a bit more, and see why a 0x201 error is generated?

guanshengliang (approved):

PR #3956, opened 2020-10-21
Added catfish extension from micli

micli:
A Hivemq-TDengine-extension that can receive data from Hivemq and save MQTT data into TDengine database.

origin repository: https://www.github.com/micli/catfish


sangshuduo (commented):
没有 实现 JDBC 接口，已和提交者沟通关闭此 PR

PR #3981, opened 2020-10-23
odbc connector, keep going

freemine:


sangshuduo:
I see the latest commit is already defect-free on the coverity scan, if the YYAC style code issues do not count.

PR #4128, opened 2020-11-05
Feature/file

jtao1735:


src/util/src/tfile.c
```diff
@@ -0,0 +1,84 @@
...
+}
+
+ssize_t tfwrite(int64_t tfd, const void *buf, size_t count) {
+
+  void *p = taosAcquireRef(tsFileRsetId, tfd);
+  if (p == NULL) return -1;
+
+  int fd = (int)(uintptr_t)p;
+
+  ssize_t ret = write(fd, buf, (uint32_t)count);
```
hzcheng: use taosTWrite() instead of the primitive write() function.

src/util/src/tfile.c
```diff
@@ -0,0 +1,84 @@
...
+}
+
+ssize_t tfread(int64_t tfd, void *buf, size_t count) {
+
+  void *p = taosAcquireRef(tsFileRsetId, tfd);
+  if (p == NULL) return -1;
+
+  int fd = (int)(uintptr_t)p;
+
+  ssize_t ret = read(fd, buf, (uint32_t)count);
```
hzcheng: use taosTRead() not read() function

PR #4297, opened 2020-11-19
fixed armv7l bus error

xiaoyanrui:
struct SStatusMsg member is not 4-byte alignment in armv7l 32bit platform


sangshuduo (changes requested):
Since adding two bytes to TSDB_EP_LEN may lead incompatible issue, please consider other way to resolve this issue.

sangshuduo:
目前没有合适设备测试验证，暂时关闭。请后续再提交对 ARM32 的支持。

PR #4459, opened 2020-12-06
[TD-2342]<fix>: disallow cross db create child table

stephenkgu:


guanshengliang:
use int32_t instead of size_ t.
for the  its length may be 4 or 8 in different platforms

stephenkgu:
int32_t cannot compile on windows, and cannot pass CI neither.

PR #4472, opened 2020-12-07
Set connection parameter with table as conventional style. 

robotspace:
1. Set connection parameter with table as conventional style.
2. Output result for test case.
3. Implement connector for OpenResty based on lua51.


sangshuduo (changes requested):
Please fix warning, update document and fix error. Thanks.

sangshuduo (approved):
Verified and looks good to me.

PR #4692, opened 2020-12-23
[TD-2502]<feature>: force version checking with rpc request messages

stephenkgu:


stephenkgu:
Better to drop connection when version mismatched.

stephenkgu:
use #4693

PR #4693, opened 2020-12-24
[TD-2502]<feature>: [v2] force version checking with rpc request messages

stephenkgu:


stephenkgu:
use #4715, just to pass buggy jenkins testing.

PR #4928, opened 2021-01-16
initiative porting to MacOSX

freemine:
this is the first trial. currently, client side tools are available, but server-side shall be easily done later.

to make original code less modified, we mock 'epoll apis' on MacOSX with the underlying kqueue system calls.

**to try as usual:**

1. git pull to your dev env.
2. cmake -B debug
3. cmake --build debug
4. cmake --install debug

**to use taos:**
./debug/build/bin/taos -c \<path to taos.cfg\> -h \<fqdn where taosd locates\>

**current issue:**
issue1: **different than FreeBSD's original kqueue implementation, MacOSX's kqueue64(...) is NOT a cancellation point, which means pthread_cancel might not work as expected if kevent64(...) is still in progress. since there's no long-blocking epoll_wait call for the current code base in taos, this issue seems no harmful for the moment.**

**TODO**: better add a epoll_cleanup for the user to cleanup internal resources (mostly caches) when terminating

### DIFF
--- a/src/os/inc/osDef.h
+++ b/src/os/inc/osDef.h
@@ -95,7 +95,7 @@ extern "C" {
   #define threadlocal __thread
 #else
   // #define threadlocal
-  #error please follow with the target platform's thread-local implementation
+  #error please follow with the thread-local implementation on the target platform 
 #endif
 
 #ifdef __cplusplus


sangshuduo (approved):
taos and taosdemo work fine now.  I suggest merging this PR then refine the detail and see feedback from the community.

sangshuduo (changes requested):
this PR breaks TDinternal build, please hold.

sangshuduo (approved):
TDinternal need support Mac too. But the community version can pick up this PC first.

PR #5076, opened 2021-01-28
taosIterateRef: shall NOT return removed node

freemine:


src/util/src/tref.c
```diff
@@ -289,52 +290,56 @@ void *taosIterateRef(int rsetId, int64_t rid) {
...
-    // rid is there
-    pNode = pNode->next;
-    if (pNode == NULL) { 
-      taosUnlockList(pSet->lockedBy+hash);
-      hash++;
+    if (pNode == NULL) {
+      for (; hash < pSet->max; ++hash) {
+        taosLockList(pSet->lockedBy+hash);
+        pNode = pSet->nodeList[hash];
+        if (pNode) break;
```
dapan1121: why not check removed here so we don't need the whole big loop?

freemine: because there are several routes to find 'next valid' pNode, the one in the the current hash-slot, the one in the next hash-slot and so on. these valid pNode in first hand shall be all checked if it's removed or not.
thus, it seems like it can not be done in just one place around line 323~324.
please correct me if there's anything wrong.

dapan1121: I mean we can check the current pNode's removed flag and loop its Next to find a node not removed, if all are removed hash++ can forward to next hash-slot. so we can add a while loop between line 325 and 326 to check all pNodes in current hash-slot, right?

freemine: well, i guess i know what you mean. the reason why i chose the original path is that i wish this PR is as less intrusive as possible to the original code base.

guanshengliang (approved):

PR #5554, opened 2021-03-24
[TD-3385]<test>: add function testcase-query stable group by general column

tomchon:
add function testcase:query stable group by general column


plum-lihui:
重新提交。

PR #5591, opened 2021-03-26
[TD-3541]<fix>: [sync] sort nodes to fix leader election failure

stephenkgu:


stephenkgu:
peerInfo sorting will cause the leader concentration issue.

PR #5738, opened 2021-04-09
Resolve problems that is met in CentOS7.

robotspace:
Set value for std with c99 to avoid compile error on CentOS7.
Get parameter from stack by same sequence.
Add performance test.


sangshuduo (commented):
LGTM

PR #5797, opened 2021-04-13
[TD-3758]<fix>: fix setnull throw exception when jdbcType is OTHER

zyyang90:
[TD-3758]<fix>: fix setnull throw exception when jdbcType is OTHER


sangshuduo (approved):
LGTM

PR #5818, opened 2021-04-14
[TD-3762]<feature>: support us and ms percision for timestamp in JDBC

zyyang90:
[TD-3762]<feature>: support us and ms percision for timestamp in JDBC


sangshuduo (approved):
LGTM

PR #5892, opened 2021-04-22
Hotfix/td 3841

zyyang90:
[TD-3841]<fix>: getFloat methed get 0 when query null value, the resultSet wasNull should be true


sangshuduo (approved):
LGTM

PR #5927, opened 2021-04-26
Hotfix/td 3962

zyyang90:
[TD-3962]<fix>: fix insert special character error in JDBC


freemine:
seems like suffered SQL injection in case when There is no timestamp string bin as para.?

sangshuduo (approved):
LGTM

PR #5940, opened 2021-04-27
[TD-3639] <fix>: compare timestamp with datetime in python

zitsen:
By this commit, It will close TD-3639.

It will not require test case to write *correct* datetime string to pass
a test case. One test could use seconds, milliseconds or microseconds to
check data.

That means the three tests are equivalent:

```python
.checkData(0, 0, "2020-01-01 00:00:00")
.checkData(0, 0, "2020-01-01 00:00:00.000")
.checkData(0, 0, "2020-01-01 00:00:00.000000")
```


sangshuduo (approved):
LGTM

PR #5945, opened 2021-04-27
[TD-3983] <enhance>: support use cfg in go connector

zitsen:
The submodule update contains TD-3983 TD-3912 fixes.


sangshuduo (approved):
LGTM

PR #6014, opened 2021-05-06
Hotfix/sangshuduo/td 3976 taosdemo print insert perf per thread for develop

sangshuduo:


reevesPAC (approved):
This commit looks good to me.

PR #6018, opened 2021-05-06
[TD-4048] <fix>: fix python connector error where ts is null

zitsen:


sangshuduo (changes requested):
the connector version in setup.py needs a bump too.

zitsen:
bump version to v2.0.9

sangshuduo (approved):
LGTM

PR #6025, opened 2021-05-07
[TD-4808] <fix>: fix python connector error where ts is null

zitsen:


zitsen:
Need merge into master from @wu-champion

PR #6036, opened 2021-05-07
[TD-182]<enhance>: use single repo for python connector

zitsen:
Remove code for each platform and build up one single python code base
for windows/osx/linux platforms and python2/python3 runtime.

### DIFF
--- a/src/connector/python/taos/connection.py
+++ b/src/connector/python/taos/connection.py
@@ -17,8 +17,7 @@ class TDengineConnection(object):
         self._config = None
         self._chandle = None
 
-        if len(kwargs) > 0:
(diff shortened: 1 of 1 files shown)


zitsen:
Test failed at here:

```
stream/showStreamExecTimeisNull.py failed

=====================log===================== 

2021-05-08 10:09:31.405220 sql: show streams except raise ValueError('year -292275055 is out of range'), actually not
```

Actually, null time will not produce error from #6018 , the test case should be reconsider

zitsen:
@wu-champion

sangshuduo (approved):
LGTM

PR #6039, opened 2021-05-08
[TD-4083]<fix>: taosdemo query times need be more than 0.

sangshuduo:


angryTOASTpac (commented):
Ok fair enough.  I do see the specific necessity to calculate and ''print ("n"),[at the end of string/timer-Label(s);. That's my bad

angryTOASTpac (commented):
Looks good

REF:src/kit/taosdemo/taosdemo.c

PR #6056, opened 2021-05-08
Initial fuzzing integration

DavidKorczynski:
Hi,

I would like to set up continuous fuzzing of TDengine by way of OSS-Fuzz. Essentially, OSS-Fuzz is a free service run by Google that performs continuous fuzzing of important open source projects. The only expectation of integrating into OSS-Fuzz is that bugs will be fixed. This is not a "hard" requirement in that no one enforces this and the main point is if bugs are not fixed then it is a waste of resources to run the fuzzers, which we would like to avoid.

In this PR https://github.com/google/oss-fuzz/pull/5675 I have done exactly that, namely created the necessary logic from an OSS-Fuzz perspective to integrate TDengine. If you would like to integrate, could you please provide a set of email(s) that will get access to the data produced by OSS-Fuzz, such as bug reports, coverage reports and more stats. The emails should be linked to a Google account in order to view the detailed reports and notice the emails affiliated with the project will be public in the OSS-Fuzz repo, as they will be part of a configuration file.

Let me know what you think

### DIFF
--- a/tests/fuzz/sql-fuzzer.c
+++ b/tests/fuzz/sql-fuzzer.c
@@ -1,0 +1,15 @@
+#include "qSqlparser.h"
+
+int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size){
+    char *new_str = (char *)malloc(size+1);
+    if (new_str == NULL){
+        return 0;
+    }
+    memcpy(new_str, data, size);
+    new_str[size] = '\0';
+
+    qSqlParse(new_str);
+
+    free(new_str);
+    return 0;
+}


guanshengliang (approved):

DavidKorczynski:
Thanks for merging it in @guanshengliang  - do you have a set of emails I can use for the OSS-Fuzz integration?

yu285:
hi friend，we are from TAOS Data ,the  corporation  that  own TDengine
thanks for your contribution .
do u mind provide us your contact information，if you have WeChat account ,that would be better .then you could join us in WeChat groups to communicate with other friends who has similar contribution to us like you.

DavidKorczynski:
I don't have a WeChat account unfortunately. Really the only thing I would need is one or more emails that can be recipients for the bug reports. Do you think we could arrange that on here or perhaps over email?

sangshuduo:
@DavidKorczynski  I am very happy to receive the bug report email. My email address is sdsang \<AT> taosdata.com. Thanks for your help.

PR #6099, opened 2021-05-12
[TD-4154]insert error msg

dapan1121:


angryTOASTpac:
@dapan1121 Ok so by inserting message in place/label according to expected "resove-t" will alleviate the 'get-error'--am I interpreting this correctly?

REF: @hjxilinx added a commit that referenced this "Merge pull Request #6099

PR #6101, opened 2021-05-12
[TD-4089]<fix>: fix systemd sent SIGKILL before taosd shutdown normally

zyyang90:
[TD-4089]<fix>: fix systemd sent SIGKILL before taosd shutdown normally


sangshuduo (approved):
LGTM

PR #6123, opened 2021-05-13
fix some type errors, and reduce potential errors

bit-hope:
fix some type errors, and add compile-instruction in order to reduce potential errors.
There exists much "declaration shadows"、"unused params"、"compare value between different types" in source code.
We can add "-Wall -Wshadow -Wextra -Werror" to check those potential errors, but we would better do not add "-Wextra" at present, because it cause so much warnings.


guanshengliang (approved):

yu285:
hi friend，we are from TAOS Data ,the  corporation  that  own TDengine
thanks for your contribution .
do u mind provide us your contact information，if you have WeChat account ,that would be better .then you could join us in WeChat groups to communicate with other friends who has similar contribution to us like you.

PR #6237, opened 2021-05-26
Feature/sangshuduo/td 4068 taosdemo stmt

sangshuduo:


angryTOASTpac:
@sangshuduo Ok I'm checking it all out now.  The file change looks good.

REF: feature/sangshuduo/TD-6048-taosdemo-stmt

PR #6355, opened 2021-06-02
[TD-1920]fix compact bug,load dnode-globalcfg mode

lichuang:
some wal compact test failed,fix the compact bug because of unload module


angryTOASTpac (approved):
OK @lichuang looks good, I'll test new-configuration soon on REVVL4+

REF:TD-1920

guanshengliang (approved):

PR #6365, opened 2021-06-03
[TD-3126]<feature>support nanoseconds time precision

shenglian-zhou:


shenglian-zhou:
please don't merge now. just for running tests

PR #6459, opened 2021-06-11
Update docs.md

Aries-Lee1991:
Update getting-started.md


Eliacy (changes requested):
There's a link targeting https://www.taosdata.com/cn, it's a Chinese version doc.

PR #6464, opened 2021-06-11
[TD-4437]<doc> fix matlab doc

xleili:
update windows Matlab connect to TDengine through taos-JDBC


sangshuduo (changes requested):
hold this PR until JDBC connect re-verified.

PR #6497, opened 2021-06-15
Add support to build queryTest, cliTest and osTest

junli1026:
1. Add support to cliTest, osTest and queryTest.

2. queryTest failed on qTsbuf.c  "assert(pDestBuf->fileSize == oldSize + size);" .  Feels like the fstat returns the old file size, I suspect this is caused by cache. So added the logic of close/reopen file, thus queryTest passed.


sangshuduo (approved):
LGTM

junli1026:
> LGTM

Thanks! Could you please help to merge ?  😄

PR #6553, opened 2021-06-19
Run unit test 'osTest' in test-all 

junli1026:
This PR modified the unit test of test-all script, changing from running 'queryTest' only to running an array of [queryTest, osTest].

### DIFF
--- a/tests/test-all.sh
+++ b/tests/test-all.sh
@@ -423,6 +423,8 @@ fi
 
 if [ "$2" != "sim" ] && [ "$2" != "python" ] && [ "$2" != "jdbc" ]  && [ "$2" != "example" ]  && [ "$1" == "full" ]; then
   echo "### run Unit tests ###"  
+  totalUnitTests=0
+  totalUnitSuccess=0
 
   stopTaosd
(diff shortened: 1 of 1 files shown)


sangshuduo:
Is it necessary to run the unit test in test-all? Most unit test programs are the developer use to check the functionality of the early development phase on purpose.

sangshuduo (commented):
Is it necessary to run the unit test in test-all? Most unit test programs are the developer use to check the functionality of the early development phase on purpose.

junli1026:
> Is it necessary to run the unit test in test-all? Most unit test programs are the developer use to check the functionality of the early development phase on purpose.

OK, but queryTest is already there in test-all. 
I think it is very project dependent about how to use unit-test. Some projects use them as part of regression test, some  use it as part of the build process. 

It is your call to decide how to use unit test,  I just think it may benifit the code quality.

It is fine if you think

sangshuduo:
OK. Let me close it. 

Put unit test in test-all or not is case by case. 

Thanks for your enthused contribution. :)

PR #6587, opened 2021-06-22
add some error handling and comment and rename file names

junli1026:
1. rename typo "semphone" to "semaphore"
2. add comment for tsched.h
3. change the function signature for taosSchedulerTask, changing from return int to void. This is because we currently don't check any return code of the function.
4. add some error handlings for fatal error.  Because the program may run into a random state for sync error, just exit the process


junli1026:
Ping 😄

sangshuduo (approved):
LGTM

PR #6593, opened 2021-06-23
WIP: [TD-4872]<fix>: fix buffer overflow in -O3 build

zitsen:


zitsen:
realpath() expands all symbolic links and resolves references to
       /./, /../ and extra '/' characters in the null-terminated string
       named by path to produce a canonicalized absolute pathname.  The
       resulting pathname is stored as a null-terminated string, up to a
       maximum of PATH_MAX bytes, in the buffer pointed to by
       resolved_path.  The resulting path will have no symbolic link,
       /./ or /../ components.

sangshuduo (approved):
LGTM

PR #6648, opened 2021-06-28
Fix SVWriteMsg memory allocation size bug

junli1026:
1. Fix SVWriteMsg memory allocation size bug.
      SVWriteMsg contains struct SWalHead(not pointer), while the mem size is 
            `int32_t size = sizeof(SVWriteMsg) + sizeof(SWalHead) + pHead->len;`
     change to: 
             `int32_t size = sizeof(SVWriteMsg) + pHead->len;`
2. Change the field name 'pHead' to 'walHead' because it is not a pointer


stephenkgu:
Nice catch. 
I have created an request for out QA team to do more backward compatibility testing. After that, we would merge it. 

Regards,
Stephen

guanshengliang (approved):

PR #6651, opened 2021-06-28
Add init/cleaup to waltest to make it work

junli1026:


src/wal/test/waltest.c
```diff
@@ -36,7 +37,7 @@ int writeToQueue(void *pVnode, void *data, int type, void *pMsg) {
 }
 
 int main(int argc, char *argv[]) {
-  char path[128] = "/home/jhtao/test/wal";
+  char path[128] = {0};
```
stephenkgu: We need a non null default value here for help (./waltest -h) printing.
Any other fixes look good to me.

junli1026: sure, will fix.

junli1026: I put "/tmp/wal" as default value in the second iteration.

stephenkgu (approved):

sangshuduo (approved):
LGTM

guanshengliang (approved):

PR #6672, opened 2021-06-29
fix cmake scripts, improve performance, and reduce compile warning

bit-hope:
1. move some cmake scripts from root CMakeLists.txt to cmake/define.inc, and fix some cmake scripts warning in MSVC.
2. try improve performance by removed unnecessary memcpy in src/util/src/tbuffer.c.
3. reduce some compile warning.


sangshuduo:
@ExplorerBellBox  thanks for your contributions. as you might be aware, the compile not passed in CI. Maybe you can check the change in your local environment first. One more suggestion, it's better to submit separate PR to solve those different issues. Thanks again.

PR #6673, opened 2021-06-29
fix cmake scripts, improve performance, and reduce compile warning

bit-hope:
1. move some cmake scripts from root CMakeLists.txt to cmake/define.inc, and fix some cmake scripts warning in MSVC.
2. try improve performance by removed unnecessary memcpy in src/util/src/tbuffer.c.
3. reduce some compile warning.
4. check (__STDC_VERSION__ >= 201112) during using static_assert().


src/util/src/tutil.c
```diff
@@ -427,13 +427,11 @@ char *taosIpStr(uint32_t ipInt) {
...
+  assert(sizeof(float) == 4);
+  return *((const float*)(pBuf));
 }
 
 FORCE_INLINE double taos_align_get_double(const char* pBuf) {
-  double dv = 0; 
-  *(int64_t*)(&dv) = *(int64_t*)pBuf;
-  return dv; 
+  assert(sizeof(double) == 8);
+  return *((const double*)(pBuf));
```
hjxilinx, line 436: This refactor may has problem while the address of the parameter "pBuf" is not aligned to 4 bytes in memory on arm32

bit-hope: 这样的话，原代码先将内存看作 int，遇到内存首地址非 4 的整数倍时，也应该会出问题；那么，此处使用 memcpy 似乎更合理

PR #6687, opened 2021-06-30
fix cmake scripts and reduce compile warnings

bit-hope:
1. move gnu compile flags "-pipe -Wshadow" from root "CMakeLists.txt" to "cmake/define.inc"; and change compile flag "/O3" to "/O2", because "/O3" is not supported by MSVC.
2. fix compile warnings in "src/util/src/tbuffer.c", by using "union" to transform integer with float and double.
3. use "memcpy()" instead of directly type convert in "src/util/src/tutil.c" to resolve memory alignment problem in ARM platform.
4. reduce other compile warnings.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,13 +42,6 @@ INCLUDE(cmake/env.inc)
 INCLUDE(cmake/version.inc)
 INCLUDE(cmake/install.inc)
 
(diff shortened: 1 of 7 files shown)


ExplorerBellBox:
I can not access continuous-integration/jenkins/pr-merge, please show the error information to me.

hjxilinx:
I have solved the compile problem and a potential bug, and your PR will pass the check soon.

PR #6796, opened 2021-07-07
Feature/td 4666 1

kailixu:
Support max cols 4096:
1. trim NULL values during generating SDataRow or SKVRow.
2. the NULL in raw payload not trimmed


kailixu:
Close as the performance problem is solved in TD-4666

PR #6862, opened 2021-07-14
Fix the problem of parsing http protocol when the request header lacks spaces

Yurunsoft:
Before the fix, this curl command is wrong

```shell
curl --location --request POST 'http://127.0.0.1:6041/rest/sql' \
--header 'Authorization:Basic cm9vdDp0YW9zZGF0YQ==' \
--header 'Content-Type:text/plain' \
--data-raw 'show databases'
```

Fix: failed to parse sp

Now support:

`Key: Value`
`Key:Value`


src/plugins/http/inc/httpParser.h
```diff
@@ -100,6 +100,7 @@ typedef enum HTTP_PARSER_STATE {
   HTTP_PARSER_CHUNK,
   HTTP_PARSER_END,
   HTTP_PARSER_ERROR,
+  HTTP_PARSER_OPTIONAL_SP,
```
winshining: It's better to remove the last comma.

winshining (approved):

src/plugins/http/inc/httpParser.h
Yurunsoft: OK, but there are many places in tdengine that end with commas.
Should we develop a common standard and introduce testing tools

winshining: Testing tools, plugins, patches and documents are all appreciated:)

PR #6936, opened 2021-07-20
[TD-5234]<feature>: added name for threads

winshining:
Now threads' names can be found in `top` (top -Hp pid).


sangshuduo (changes requested):
something wrong lead connectors changed? go, grafanaplugin, hivemq...

PR #6968, opened 2021-07-22
[TD-5454]<fix>: fix arm32 qDebug error

afwerar:
change PRIx64 to %p——The void * pointer argument is printed in hexadecimal (as if by %#x or %#lx)。


sangshuduo (approved):
LGTM

hjxilinx (approved):

PR #6989, opened 2021-07-23
[TD-5489]<fix>: use jdbc specified version number in cmake/CMakeList.…

zyyang90:
[TD-5489]<fix>: use jdbc specified version number in cmake/CMakeList.txt and src/connector/jdbc/CMakeList.txt


sangshuduo (approved):

sangshuduo (approved):
LGTM

PR #7072, opened 2021-07-29
[TD-5631]<fix>fix functionCompatList array overflow access

lichuang:


shenglian-zhou (commented):
during block info merge, accumulate small blocks

shenglian-zhou (approved):
if possible, could a test case be added?

hjxilinx (approved):

PR #7097, opened 2021-07-31
Feature/td 5221 - prepare/non-prepare：优化taos payload拼接方式

kailixu:
TD-5221 - prepare/non-prepare：优化taos payload拼接方式


kailixu:
close temporarily to avoid triggering CI

PR #7113, opened 2021-08-02
[TD-5578]<fix> the max length of like condition can be configured

wangmm0220:


src/common/src/tglobal.c
```diff
@@ -984,6 +986,16 @@ static void doInitGlobalConfig(void) {
   cfg.unitType = TAOS_CFG_UTYPE_BYTE;
   taosInitConfigOption(cfg);
 
+  cfg.option = "maxLikeLength";
```
hjxilinx: wildcardstringlength would be better

PR #7144, opened 2021-08-04
Fix dirname bug on MacOS

junli1026:
1. Fix bug dirname() causing bug when creating dirs on MaxOS.
The fix is to make a copy of dirname(s) because the implementation of 'dirname' differs on different platforms. Some platform may modify the contents of the string passed into dirname(). Others may return a pointer to internal static storage space that will be overwritten by next call. For case like that, we should not use the pointer directly.
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dirname.3.html

2. the memory located by strdup() should be freed by free(), instead of internal function tfree().

### DIFF
--- a/src/tfs/src/tfs.c
+++ b/src/tfs/src/tfs.c
@@ -261,11 +261,20 @@ int tfsMkdirRecurAt(const char *rname, int level, int id) {
       // Try to create upper
       char *s = strdup(rname);
 
-      if (tfsMkdirRecurAt(dirname(s), level, id) < 0) {
-        tfree(s);
+      // Make a copy of dirname(s) because the implementation of 'dirname' differs on different platforms.
+      // Some platform may modify the contents of the string passed into dirname(). Others may return a pointer to
+      // internal static storage space that will be overwritten by next call. For case like that, we should not use
+      // the pointer directly in this recursion.
+      // See https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dirname.3.html
+      char *dir = strdup(dirname(s));
+
+      if (tfsMkdirRecurAt(dir, level, id) < 0) {
+        free(s);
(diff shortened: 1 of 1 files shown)


glzhao89 (approved):

src/tfs/src/tfs.c
```diff
@@ -261,11 +261,20 @@ int tfsMkdirRecurAt(const char *rname, int level, int id) {
...
       char *s = strdup(rname);
 
-      if (tfsMkdirRecurAt(dirname(s), level, id) < 0) {
-        tfree(s);
+      // Make a copy of dirname(s) because the implementation of 'dirname' differs on different platforms.
+      // Some platform may modify the contents of the string passed into dirname(). Others may return a pointer to
+      // internal static storage space that will be overwritten by next call. For case like that, we should not use
+      // the pointer directly in this recursion.
+      // See https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dirname.3.html
+      char *dir = strdup(dirname(s));
```
kailixu, line 269: 1. The parameter 's' in dirname(s) is already a duplicated string of 'rname'. In this scenario, there is no other side effects even 's' is modified. 
2. Seems both free and tfree works. Why free() is a must?

src/tfs/src/tfs.c
```diff
@@ -261,11 +261,20 @@ int tfsMkdirRecurAt(const char *rname, int level, int id) {
...
+
+      if (tfsMkdirRecurAt(dir, level, id) < 0) {
+        free(s);
+        free(dir);
         return -1;
       }
-      tfree(s);
+      free(s);
```
kailixu, line 276: Why free() is a must? Seems both free and tfree works normally here.

guanshengliang (approved):

PR #7150, opened 2021-08-04
[TD-5725]<fix>: taosdump --help, regarding acceptable timestamp.

sangshuduo:


sangshuduo:
TDengine will support three cases. close this one.

PR #7152, opened 2021-08-04
[TD-5712]<fix>: taosdump timestamp overflow.

sangshuduo:


sangshuduo:
TDengine will support three cases. close this one.

PR #7164, opened 2021-08-04
Fix/td 5788

zyyang90:
[TD-5788]<fix>: fix DatabaseMetaData's ResultSet with wrong taos type


sangshuduo (commented):
should we use meaningful name rather than col6, col7...?

sangshuduo (approved):

PR #7176, opened 2021-08-05
enhance: improve inserting logic, handle memory allocation error

cosmic-jc:


cosmic-jc:
Memory allocation error handling is still not finished because the failure in upper level is complicated and cannot rollback easily

cosmic-jc:
The memory allocation logic and upper usage needs to be reviewed and refactored, for it is impossible that there are so much places to fail. It is ugly.

cosmic-jc:
In tsdbRead, forcing set null value is necessary because of some logic leak. Need to refactor.

hzcheng (approved):

kailixu (approved):

PR #7177, opened 2021-08-05
Hotfix/td 5765: check max length when alter tag value.

kailixu:
[TD-5765]<hotfix>:check max length when alter tag value.


kailixu:
Close as to decrease importing other modifications.

PR #7291, opened 2021-08-10
Feature/ Extract schema file

cosmic-jc:


cosmic-jc:
TODO：
1. handle sync
2. after first upgrading, dump meta to schema and compact meta file

PR #7443, opened 2021-08-18
[TD-6199]<fix>fix sdb mnode data memory leak

lichuang:


src/mnode/src/mnodeSdb.c
```diff
@@ -549,6 +549,14 @@ static int32_t sdbInsertHash(SSdbTable *pTable, SSdbRow *pRow) {
   }
 
   pthread_mutex_lock(&pTable->mutex);
+  void **ppRow = (void **)taosHashGet(pTable->iHandle, key, keySize);
+  if (ppRow) {
+    SSdbRow row = {
```
hzcheng, line 554: I prefer a test case that can prove the memory leak and validate this PR. How about getting a test case to prove the problem at first?

cosmic-jc: Please refer to line 282 in src/util/src/hash.c

PR #7484, opened 2021-08-20
change version numbe ver 2.0.20.15

tomchon:


tomchon:
this is hotfix for https://jira.taosdata.com:18080/browse/TD-6230

PR #7523, opened 2021-08-23
[TD-6145]<feature>:tagfilter-add parser support

shenglian-zhou:


src/client/src/tscSQLParser.c
```diff
@@ -4486,6 +4499,44 @@ static int32_t validateLikeExpr(tSqlExpr* pExpr, STableMeta* pTableMeta, int32_t
...
+      return invalidOperationMsg(msgBuf, msg2);
+    }
+
+    int errCode = 0;
+    regex_t regex;
+    char    regErrBuf[256] = {0};
+
+    const char* pattern = pRight->value.pz;
+    int cflags = REG_EXTENDED | REG_ICASE;
+    if ((errCode = regcomp(&regex, pattern, cflags)) != 0) {
```
dapan1121, line 4529: do we need to free regex?

src/util/src/tcompare.c
```diff
@@ -344,6 +350,43 @@ int32_t compareStrPatternComp(const void* pLeft, const void* pRight) {
...
+    regfree(&regex);
+    free(str);
+    free(pattern);
+    return 1;
+  }
+
+  errCode = regexec(&regex, str, 0, NULL, 0);
+  if (errCode != 0 && errCode != REG_NOMATCH) {
+    regerror(errCode, &regex, msgbuf, sizeof(msgbuf));
+    uError("Failed to match %s with pattern %s, reason %s", str, pattern, msgbuf)
```
dapan1121: 不匹配也正常吧，不需要错误日志吧

src/client/src/tscSQLParser.c
shenglian-zhou: It will be freed.

src/util/src/tcompare.c
shenglian-zhou: yes. I will change it to udebug.

PR #7653, opened 2021-08-27
[TD-6403]<fix>: merge python connector 2.1.0 to master

zitsen:


src/connector/python/README.md
```diff
@@ -11,8 +12,417 @@ pip install ./TDengine/src/connector/python
...
+conn.insert_lines(lines)
+
+result = conn.query("show tables")
+for row in result:
+    print(row)
+result.close()
+
+
+conn.exec("drop database if exists %s" % dbname)
+conn.close()
```
sangshuduo, line 422: wondering why put so large examples code block in README. Especially they are also in py files.

src/connector/python/taos/__init__.py
```diff
@@ -1,20 +1,478 @@
...
+
+
+conn.exec("drop database if exists %s" % dbname)
+conn.close()
+
+```
+
+## License - AGPL-3.0
+
+Keep same with [TDengine](https://github.com/taosdata/TDengine).
```
sangshuduo, line 430: same question as previous

src/connector/python/taos/bind.py
```diff
@@ -0,0 +1,437 @@
...
+                ts = int(round((value - _datetime_epoch).total_seconds() * 1000))
+            elif precision == PrecisionEnum.Microseconds:
+                ts = int(round((value - _datetime_epoch).total_seconds() * 10000000))
+            else:
+                raise PrecisionError("datetime do not support nanosecond precision")
+        elif type(value) is float:
+            if precision == PrecisionEnum.Milliseconds:
+                ts = int(round(value * 1000))
+            elif precision == PrecisionEnum.Microseconds:
+                ts = int(round(value * 10000000))
```
sangshuduo, line 94: is the INT range enough to present all possible values?

sangshuduo: which that mean we don't support Python2 anymore?

src/connector/python/taos/cinterface.py
```diff
@@ -1,295 +1,839 @@
...
+    _libtaos.taos_cleanup()
+
+
+_libtaos.taos_get_client_info.restype = c_char_p
+
+
+def taos_get_client_info():
+    # type: () -> str
+    """Get client version info.
+    获取客户端版本信息。
```
sangshuduo, line 106: prefer only English comments

src/connector/python/taos/cinterface.py
```diff
@@ -1,295 +1,839 @@
...
+
+
+def taos_open_stream(connection, sql, callback, stime=0, param=None, callback2=None):
+    # type: (ctypes.c_void_p, str, stream_callback_type, c_int64, c_void_p, c_void_p) -> ctypes.pointer
+    if callback2 != None:
+        callback2 = stream_callback2_type(callback2)
+    """Open an stream"""
+    return c_void_p(
+        _libtaos.taos_open_stream(
+            connection, ctypes.c_char_p(sql.encode("utf-8")), stream_callback_type(callback), stime, param, callback2
```
sangshuduo, line 576: why callback2 is needed?

src/connector/python/README.md
zitsen: People will see it in PyPI, https://pypi.org/project/taospy/

src/connector/python/taos/bind.py
zitsen: 1. Python use 64 bit int in x64 arch.
2. Both python2/3 are supported.

src/connector/python/taos/cinterface.py
zitsen: It's taos.h defines like.

```c
DLL_EXPORT TAOS_STREAM *taos_open_stream(TAOS *taos, const char *sql, void (*fp)(void *param, TAOS_RES *, TAOS_ROW row),
                              int64_t stime, void *param, void (*callback)(void *));
```

src/connector/python/taos/cinterface.py
zitsen: I agree, I'll open another pr for both master/develop. But may I keep it the same with current develop version now?
Master release deadline  is 8.30 as required by @shenglian-zhou .

sangshuduo:
merged.

zitsen:
Thanks. Document improvement is tracking with https://jira.taosdata.com:18080/browse/TD-6436 now.

PR #7711, opened 2021-08-30
[TD-6449]<doc>: prefer english only in method documentations in python connector

zitsen:
I must mention that the commit has been desclared as `[ci skip]` in commit message.


sangshuduo (approved):
LGTM

PR #7732, opened 2021-08-31
[TD-6169]<fix>: windows dll client can not quit.

afwerar:


sangshuduo (approved):
LGTM

src/util/src/tcache.c
```diff
@@ -60,6 +60,8 @@ static FORCE_INLINE void __cache_lock_destroy(SCacheObj *pCacheObj) {
  */
 static void doCleanupDataCache(SCacheObj *pCacheObj);
 
+void taosStopCacheRefreshWorker(void);
+
```
wangmm0220: Is this function define duplicated ? It is already defined in tcache.h.

wangmm0220 (approved):

src/util/src/tcache.c
afwerar: It has been removed in the next commit.

PR #7763, opened 2021-09-02
[TD-6442]<feature>: Support OpenTSDB telnet style data import format

glzhao89:


src/client/src/tscParseOpenTSDB.c
```diff
@@ -0,0 +1,423 @@
...
+}
+
+int taos_insert_telnet_lines(TAOS* taos, char* lines[], int numLines) {
+  int32_t code = 0;
+
+  SSmlLinesInfo* info = tcalloc(1, sizeof(SSmlLinesInfo));
+  info->id = genUID();
+
+  if (numLines <= 0 || numLines > 65536) {
+    tscError("OTD:0x%"PRIx64" taos_insert_telnet_lines numLines should be between 1 and 65536. numLines: %d", info->id, numLines);
```
dapan1121, line 368: free info

src/client/src/tscParseOpenTSDB.c
```diff
@@ -0,0 +1,423 @@
...
+    cur++;
+    len++;
+  }
+  if (len == 0) {
+    tfree(pSml->stableName);
+    return TSDB_CODE_TSC_LINE_SYNTAX_ERROR;
+  }
+
+  pSml->stableName[len] = '\0';
+  *index = cur + 1;
```
dapan1121, line 67: need to check if *cur is '\0' now.

src/client/src/tscParseOpenTSDB.c
```diff
@@ -0,0 +1,423 @@
...
+    if (ret) {
+      tscError("OTD:0x%"PRIx64" Unable to parse key", info->id);
+      return ret;
+    }
+    ret = parseTelnetTagValue(pkv, &cur, &is_last_kv, info);
+    if (ret) {
+      tscError("OTD:0x%"PRIx64" Unable to parse value", info->id);
+      return ret;
+    }
+    if ((strcasecmp(pkv->key, "ID") == 0) && pkv->type == TSDB_DATA_TYPE_BINARY) {
```
dapan1121, line 265: only one ID allowed, right? if so, we need to add more check.

src/client/src/tscParseOpenTSDB.c
glzhao89: Done. Added check if *cur is '\0' return error since we expect more fields in the string. Also added same check logic for metric, timestamp, metric value and tag key, as these should not be the last fields in string(except for tag value).

src/client/src/tscParseOpenTSDB.c
glzhao89: Done.

PR #7863, opened 2021-09-10
[TS-337]<test>: add test case for interp function returns empty result

xiaopingcs:
add test case for interp function returns empty result


liuyq-617 (approved):
LGTM

PR #7896, opened 2021-09-13
[TD-5757]<fix>: taosdump precision.

sangshuduo:


wenzhouwww (commented):
LGTM

wenzhouwww (commented):
The test case contains the interpretation of the database median value after adaptive accuracy！

PR #7904, opened 2021-09-14
[TD-6633]<test>:modify python file

happyguoxy:
modify python file


sangshuduo (commented):
it will fail when Sep/1 2022 coming.

Issue #7844, opened 2021-09-09
The image typesetting of the English document is incorrect


jiacy-jcy:
![图片](https://user-images.githubusercontent.com/57592088/132648504-cdc8d7aa-be34-4f41-88ce-d8cd385121ba.png)
Here is the Chinese documentation picture,pitcure 3 is the TDengine Master Writing Process.But in English documentation,image cannot be displayed.
As a result of this problem, other images in the document also have errors.
We hope someone can help us fix this problem.
Specific repair methods are as follows: According to Chinese documentation ,modify the English documentation.Make the illustrations in the two documents consistent.
Here is the link of Chinese document:https://github.com/taosdata/TDengine/blob/develop/documentation20/cn/03.architecture/docs.md
Here is the link of English document:https://github.com/taosdata/TDengine/blob/develop/documentation20/en/03.architecture/docs.md

And the link of picture maybe useful:https://github.com/taosdata/TDengine/blob/develop/documentation20/en/images/architecture/write_master.png
https://github.com/taosdata/TDengine/blob/develop/documentation20/en/images/architecture/write_slave.png

daaaaaaaaaaa:
It seems that the pictures of Chinese documents cannot be displayed?

daaaaaaaaaaa:
pick up challenge

Aries-Lee1991:
> pick up challenge

Assigned it to you.

Aries-Lee1991:
> It seems that the pictures of Chinese documents cannot be displayed?
> ![中文文档无法显示图片](https://user-images.githubusercontent.com/51794124/132951663-40e7d919-6041-4164-909d-dfe8a68927a7.PNG)

And the link of picture maybe useful:https://github.com/taosdata/TDengine/blob/develop/documentation20/en/images/architecture/write_master.png
https://github.com/taosdata/TDengine/blob/develop/documentation20/en/images/architecture/write_slave.png

Aries-Lee1991:
@daaaaaaaaaaa We will make a certificate for the new contributor and give a souvenir. Can you leave your contact information or add my wechat (690754649).

PR #7924, opened 2021-09-14
Solve the problem of picture typesetting in English and Chinese documents.

daaaaaaaaaaa:
good first issue：https://github.com/taosdata/TDengine/issues/7844

### DIFF
--- a/documentation20/en/03.architecture/docs.md
+++ b/documentation20/en/03.architecture/docs.md
@@ -155,7 +155,7 @@ The design of TDengine is based on the assumption that one single node or softwa
 
 Logical structure diagram of TDengine distributed architecture as following:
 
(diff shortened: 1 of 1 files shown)


jiacy-jcy (approved):

jiacy-jcy (commented):
The url of the image should be 'page: // ' but not 'https: //'，such cases are not mistakes .
The pictures in the Chinese document are correct, you need to modify the English document according to the Chinese document.

daaaaaaaaaaa:
> The url of the image should be 'page: // ' but not 'https: //'，such cases are not mistakes .
> The pictures in the Chinese document are correct, you need to modify the English document according to the Chinese document.

But it seems that the pictures of Chinese documents cannot be displayed?
图片不能显示

Aries-Lee1991 (commented):
@Eliacy

Aries-Lee1991 (commented):
@Eliacy

Eliacy:
LGTM

Issue #7857, opened 2021-09-10
Is anybody interested to help us to modify the documentation?


zhaoyanggh:
There are some connectors missing in The Efficient Data Querying page, with both Chinese and English. Links are: 
https://github.com/taosdata/TDengine/blob/develop/documentation20/cn/06.queries/docs.md and
https://github.com/taosdata/TDengine/blob/develop/documentation20/en/06.queries/docs.md
![Screen Shot 2021-09-10 at 9 42 07 AM](https://user-images.githubusercontent.com/47117543/132785282-5759914d-e61c-46c8-a8f8-1c5169bc2a2c.png)
To find the missing connectors, you can check the Efficient Data Writing page, with the links:
https://github.com/taosdata/TDengine/blob/develop/documentation20/cn/05.insert/docs.md or
https://github.com/taosdata/TDengine/blob/develop/documentation20/en/05.insert/docs.md

Yonghui-Lee:
pick up challenge

Aries-Lee1991:
> pick up challenge

Assigned it to you.

Aries-Lee1991:
@Yonghui-Lee We will make a certificate for the new contributor and give a souvenir. Can you leave your contact information or add my wechat (690754649).

PR #7928, opened 2021-09-15
[TD-7857]<doc>:add missing connectors

Yonghui-Lee:
link to #7857


zhaoyanggh (commented):
Actually, we also support the Nodejs connector, it would be better if you can also put that on.

Yonghui-Lee:
Add `Nodejs Connector` in the last commit

zhaoyanggh (approved):

zhaoyanggh:
LGTM

PR #7944, opened 2021-09-15
Hotfix/td 6636

kailixu:
<TD-6636><hotfix>: row convert fix and schema version fetch optimize


kailixu:
Cancel as submitted to master.

PR #7965, opened 2021-09-16
[TD-6085][TD-6314]<test>:add testcase

jiacy-jcy:
<test>add test case


xiaopingcs:
All c examples should be added into CI, please update test-all.sh file.

Issue #7991, opened 2021-09-17
Could someone modify the display issues as pictures


jiajingbin:
![image](https://user-images.githubusercontent.com/39030567/133781851-cf3bac52-e4fc-4fc7-b653-15cc6788ce6b.png)

![image](https://user-images.githubusercontent.com/39030567/133781557-63adaa1d-0ba6-4606-bc9b-47e10e9ccdb4.png)

jiajingbin:
Link----->https://www.taosdata.com/cn/documentation/administrator

BugCui:
pick up challenge

Aries-Lee1991:
> pick up challenge

Assigned it to you.

BugCui:
modify the display issues #8035

Aries-Lee1991:
@BugCui We will make a certificate for the new contributor and give a souvenir. Can you leave your contact information or add my wechat (690754649).

PR #8035, opened 2021-09-22
modify the display issues

BugCui:
modify the display issues https://github.com/taosdata/TDengine/issues/7991


jiajingbin:
**Suggestion**
Whether is it more fluent as follows?
默认值：动态获取当前客户端运行系统所在的时区。

BugCui:
> **Suggestion**
> Whether is it more fluent as follows?
> 默认值：动态获取当前客户端运行系统所在的时区。

yes,I think so.

PR #8057, opened 2021-09-24
Hotfix/sangshuduo/td 6280 taosdump p for develop

sangshuduo:


src/kit/taosdump/taosdump.c
```diff
@@ -483,7 +483,14 @@ static error_t parse_opt(int key, char *arg, struct argp_state *state) {
                 errorPrintReqArg2("taosdump", "P");
                 exit(EXIT_FAILURE);
             }
-            g_args.port = atoi(arg);
+
+            uint64_t port = atoi(arg);
+            if (port > 655325) {
```
zhaoyanggh: Should the maximum port number be 65535?

sangshuduo: good catch!

PR #8060, opened 2021-09-24
[TD-7581]<test> :" add test case about taoshell prase params by taos -s"

wenzhouwww:
add test case for taoshell execute sql by taos -s, include "taos -d test -k 1 " and "taos -D ./"


wu-champion (changes requested):
1. need add  counterexample ；
2. some programs need to add verification procedures after running ，e.g. after part "test other file name about tables.sql"

PR #8103, opened 2021-09-28
[TD-10436]<test>:add test case into CI

jiacy-jcy:


wenzhouwww (commented):
You should think about four kind of ways to give params to taosdemo 

check params -b can work and work well , even  should think about some illegal use way.

PR #8105, opened 2021-09-28
remove check client tag bytes len must over db bytes len

DuanKuanJun:


kailixu (approved):
The affect is some client would fetch truncated tag values without notification.

hjxilinx (approved):

stephenkgu (approved):

PR #8209, opened 2021-10-12
fix the wrong spelling of the word

Oaiqiy:


Aries-Lee1991:
@Oaiqiy We will make a certificate for the new contributor and give a souvenir. Can you leave your contact information or add my wechat.
image

PR #8313, opened 2021-10-19
[TD-10689]<enhance>optimize memory using for arm platform

zhaoyanggh:


src/kit/taosdemo/taosdemo.c
```diff
@@ -6400,9 +6402,12 @@ static bool getInfoFromJsonFile(char* file) {
     }
 
     if (INSERT_TEST == g_args.test_mode) {
+        memset(&g_Dbs, 0, sizeof(SDbs));
```
sangshuduo, line 6433: why need memset() if already use calloc()?

src/kit/taosdemo/taosdemo.c
```diff
@@ -5801,6 +5832,8 @@ static bool getMetaFromInsertJsonFile(cJSON* root) {
                     goto PARSE_OVER;
                 }
                 g_Dbs.db[i].superTbls[j].insertRows = insertRows->valueint;
+                if (g_Dbs.db[i].superTbls[j].insertRows > g_args.prepared_rand) {
+                }
```
sangshuduo: if (true) then what?

src/kit/taosdemo/taosdemo.c
```diff
@@ -6491,6 +6529,8 @@ static void postFreeResource() {
         }
     }
     tmfree(g_sampleBindBatchArray);
+
+    tmfree(&g_Dbs);
```
sangshuduo: free memory of heap is enough. no need free global variable.

src/kit/taosdemo/taosdemo.c
```diff
@@ -11975,7 +11992,7 @@ static void setParaFromArg() {
     tstrncpy(g_Dbs.resultFile, g_args.output_file, MAX_FILE_NAME_LEN);
 
     g_Dbs.use_metric = g_args.use_metric;
-
+    g_args.prepared_rand = g_args.insertRows;
```
sangshuduo: if the user specifies a very large number of insert rows will lead to out-of-memory. how about

#define MAX_PREPARED   10000
g_args.prepred_rand = min(g_args.insertRows, MAX_PREPARED_RAND)

zhaoyanggh: changed

src/kit/taosdemo/taosdemo.c
zhaoyanggh: removed

src/kit/taosdemo/taosdemo.c
```diff
@@ -666,6 +666,7 @@ SArguments g_args = {
     0,               // insert_interval
     DEFAULT_TIMESTAMP_STEP, // timestamp_step
     1,               // query_times
+    10000,           // prepared_rand
```
sangshuduo, line 669: use MAX_PREPARED_RAND here?

PR #8359, opened 2021-10-20
Update tsdbFile.c

TOMsworkspace:
Null pointer terminated.

### DIFF
--- a/src/tsdb/src/tsdbFile.c
+++ b/src/tsdb/src/tsdbFile.c
@@ -77,11 +77,14 @@ int tsdbEncodeSMFileEx(void **buf, SMFile *pMFile) {
 
 void *tsdbDecodeSMFileEx(void *buf, SMFile *pMFile) {
   
-  ASSERT(buf == NULL && pMFile == NULL);
+  ASSERT(pMFile == NULL);
   
   char *aname;
   buf = tsdbDecodeMFInfo(buf, &(pMFile->info));
   buf = taosDecodeString(buf, &aname);
+    
+  ASSERT(strlen(aname) < TSDB_FILENAME_LEN);
+    
   strncpy(TSDB_FILE_FULL_NAME(pMFile), aname, TSDB_FILENAME_LEN);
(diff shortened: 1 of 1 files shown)


src/tsdb/src/tsdbFile.c
```diff
@@ -76,9 +76,15 @@ int tsdbEncodeSMFileEx(void **buf, SMFile *pMFile) {
 }
 
 void *tsdbDecodeSMFileEx(void *buf, SMFile *pMFile) {
+  
+  ASSERT(pMFile == NULL);
```
kailixu, line 80: 1. Not a must check. 
2. If check, should be != NULL.

src/tsdb/src/tsdbFile.c
```diff
@@ -341,10 +347,16 @@ static int tsdbEncodeSDFileEx(void **buf, SDFile *pDFile) {
 }
 
 static void *tsdbDecodeSDFileEx(void *buf, SDFile *pDFile) {
+  
+  ASSERT(pDFile == NULL);
+
```
kailixu, line 352: 1. Not a must check. 
2. If check, should be != NULL.

src/tsdb/src/tsdbFile.c
```diff
@@ -76,9 +76,15 @@ int tsdbEncodeSMFileEx(void **buf, SMFile *pMFile) {
...
 void *tsdbDecodeSMFileEx(void *buf, SMFile *pMFile) {
+  
+  ASSERT(pMFile == NULL);
+  
   char *aname;
   buf = tsdbDecodeMFInfo(buf, &(pMFile->info));
   buf = taosDecodeString(buf, &aname);
+    
+  ASSERT(strlen(aname) < TSDB_FILENAME_LEN);
+
```
wangmm0220, line 87: aname is read from buf. The length is uncertainty. So we should not assert strlen(aname) < TSDB_FILENAME_LEN. We should control the length of strncpy.

PR #8381, opened 2021-10-22
[TD-10689]<improve>optimize memory using for arm platform(master)

zhaoyanggh:


src/kit/taosdemo/taosdemo.c
```diff
@@ -12075,8 +12235,6 @@ int main(int argc, char *argv[]) {
 
     if (g_args.metaFile) {
         g_totalChildTables = 0;
-        initOfInsertMeta();
```
sangshuduo, line 12103: remove these two functions will lead to RESTful fails

src/kit/taosdemo/taosdemo.c
```diff
@@ -3500,7 +3500,13 @@ static int postProceSql(char *host, uint16_t port,
 
     char resEncodingChunk[] = "Encoding: chunked";
     char resHttp[] = "HTTP/1.1 ";
+<<<<<<< HEAD
```
sangshuduo: need choose which part be used.

PR #8430, opened 2021-10-26
[TD-10771]<fix>: fix invalid syntax error for python2

zitsen:
Wait for @zyyang-taosdata to confirm the fix.

### DIFF
--- a/tests/examples/python/taosdemo/taosdemo.py
+++ b/tests/examples/python/taosdemo/taosdemo.py
@@ -21,78 +21,91 @@ import json
 import random
 import time
 import datetime
+import multiprocessing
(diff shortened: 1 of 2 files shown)


tests/examples/python/taosdemo/taosdemo.py
```diff
@@ -21,78 +21,91 @@
 import random
 import time
 import datetime
+import multiprocessing
```
sangshuduo, line 24: redundant?

zitsen: `# type: (multiprocessing._LOCK_TYPE` requires

PR #8488, opened 2021-10-29
fix error tips because of overflowing of convert time precision & inv…

wangmm0220:
…alidate tips in function of derivative

### DIFF
--- a/src/query/src/sql.c
+++ b/src/query/src/sql.c
@@ -25,8 +25,6 @@
 ** input grammar file:
 */
 /************ Begin %include sections from the grammar ************************/
-#line 23 "sql.y"
-
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -38,7 +36,6 @@
 #include "ttokendef.h"
 #include "tutil.h"
 #include "tvariant.h"
-#line 42 "sql.c"
 /**************** End of %include directives **********************************/
 /* These constants specify the various numeric values for terminal symbols.
 ***************** Begin token definitions *************************************/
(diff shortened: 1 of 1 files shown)


dapan1121 (approved):

src/os/src/detail/osTime.c
```diff
@@ -379,26 +386,45 @@ int64_t convertTimePrecision(int64_t time, int32_t fromPrecision, int32_t toPrec
       return time;  // only to pass windows compilation
     }
   } //end switch fromPrecision
+end_:
+  if (tempResult >= (double)INT64_MAX) return INT64_MAX;
```
shenglian-zhou, line 390: since int64_t to double conversion will lose precision, tempResult is less than INT64_MAX, greater than 9223372036854775295, the line 390 will return true.
same for int64_t near INT64_MIN
~

src/query/src/sql.c
```diff
@@ -22,9 +24,8 @@
 ** The following is the concatenation of all %include directives from the
 ** input grammar file:
 */
-#include <stdio.h>
-#include <assert.h>
 /************ Begin %include sections from the grammar ************************/
+#line 23 "sql.y"
```
shenglian-zhou: use lemon -l to remove location information from generated file?

src/os/src/detail/osTime.c
wangmm0220: It does not matter. If tempResult is enough, returning INT64_MAX is ok.

src/query/src/sql.c
wangmm0220: ok, I will see it the next time.

wangmm0220: ok. I will generate sql.c again

shenglian-zhou (approved):

PR #8517, opened 2021-11-01
add a build option to disable embedded lua lib. by default enable

robotspace:
Add a macro to disable embedded lualib because of conflicts with Lua connector.


cmake/define.inc
```diff
@@ -150,6 +150,14 @@ IF (TD_BUILD_HTTP)
   ADD_DEFINITIONS(-DHTTP_EMBEDDED)
 ENDIF ()
 
+IF (${BUILD_LUA} MATCHES "true")
```
sangshuduo: as we talked about, we need to make LUA working by default for UDF functionality.

sangshuduo (changes requested):

sangshuduo (approved):
LGTM

PR #8570, opened 2021-11-04
[TD-10773]<enhance>: add user configurable option smlChildTableName to specify child table name instead using ID in tag field.

glzhao89:


jiajingbin (approved):
LGTM

PR #8583, opened 2021-11-05
[TS-572]<fix>: fix noHttpResponseException in httpClient

zyyang90:
[TS-572]<fix>: fix noHttpResponseException in httpClient


huolibo (commented):
suggest poolUtil 's constructor is private

sangshuduo (approved):

PR #8590, opened 2021-11-05
[td-10748]interp

dapan1121:


src/query/src/qExecutor.c
```diff
@@ -6047,127 +5910,477 @@ static SSDataBlock* doIntervalAgg(void* param, bool* newgroup) {
...
+      }
+    }
+
     STableQueryInfo* pTableQueryInfo = pRuntimeEnv->current;
 
-    setTagValue(pOperator, pTableQueryInfo->pTable, pIntervalInfo->pCtx, pOperator->numOfOutput);
-    setInputDataBlock(pOperator, pIntervalInfo->pCtx, pBlock, pQueryAttr->order.order);
-    setIntervalQueryRange(pRuntimeEnv, &pBlock->info.window, pBlock->info.tid);
+    if (pEveryInfo->groupDone && pTableQueryInfo->groupIndex == pEveryInfo->lastGroupIdx) {
+      assert(pOperator->upstream[0]->notify == NULL);
```
shenglian-zhou, line 6366: did not find the location where notify is set to NULL

src/inc/taosmsg.h
```diff
@@ -479,6 +479,7 @@ typedef struct {
   bool        stateWindow;       // state window flag 
 
   STimeWindow window;
+  STimeWindow range;
```
hjxilinx: Could you please add a little bit more comment on this attribute to explain why this attribute is needed?

hjxilinx (changes requested):

src/inc/taosmsg.h
dapan1121: ok

hjxilinx (approved):
I'm OK with this PR.

src/query/src/qFill.c
```diff
@@ -493,12 +494,20 @@ int64_t getNumOfResultsAfterFillGap(SFillInfo* pFillInfo, TSKEY ekey, int32_t ma
   return (numOfRes > maxNumOfRows) ? maxNumOfRows : numOfRes;
 }
 
-int32_t taosGetLinearInterpolationVal(SPoint* point, int32_t outputType, SPoint* point1, SPoint* point2, int32_t inputType) {
-  double v1 = -1, v2 = -1;
+int32_t taosGetLinearInterpolationVal(SPoint* point, int32_t outputType, SPoint* point1, SPoint* point2, int32_t inputType, bool *exceedMax, bool *exceedMin) {
+  double v1 = -1, v2 = -1, vmax = -1, vmin = -1;
   GET_TYPED_DATA(v1, double, inputType, point1->val);
```
shenglian-zhou, line 499: for integer values, keep double calculation at the last step and as much as integer calculation as possible?

dapan1121: 好像意义不是很大，既然运算过程中肯定要 double保持精度，那就必然有需要转到double的部分。

PR #8617, opened 2021-11-09
[TS-486]<fix>: fix balance notify with replica altering

stephenkgu:


guanshengliang (approved):

guanshengliang:
When alter the database, we need to verify whether to start the balance, and there is no problem with this PR

PR #8634, opened 2021-11-10
insert datas contains multiple timestamp columns, but this tool transfer it to 'char' and pops up errors.

t00350320:
insert datas contains multiple timestamp columns, but this tool transfer it to 'char' and pops up errors.

### DIFF
--- a/importSampleData/app/main.go
+++ b/importSampleData/app/main.go
@@ -709,7 +709,7 @@ func executeBatchInsert(insertSql string, connection *sql.DB) int64 {
 }
 
 func getFieldValue(fieldValue interface{},fieldtype interface{}) string {
-	if fieldtype == "timestamp" {
+	if fieldtype == "timestamp" || fieldtype == "bigint"  {
 		return fmt.Sprintf("%v", fieldValue)
 	}
 	return fmt.Sprintf("'%v'", fieldValue)
--- a/importSampleData/config/cfg.toml
+++ b/importSampleData/config/cfg.toml
@@ -18,6 +18,8 @@ tags = [
 
 fields = [
     # 字段列表，name 为字段名称，type 为字段类型
+	# 除主键外，其他field如果也要设置为timestamp，可以是type ="timestamp" 类型，此时value可同时支持'2006-01-02 15:04:05.000'和millisecond格式
+	# 也可以是type = "bigint",此时value只支持millisecond格式
     { name = "ts",                  type = "timestamp" },
     { name = "temperature",         type = "int" },
     { name = "humidity",            type = "float" },


sangshuduo (commented):
thanks for your PR.
The valid timestamp value can be either human-readable like '2021-10-10 ...' or integer like 150000000000. I think we need to process both of them too.

t00350320:
> thanks for your PR. The valid timestamp value can be either human-readable like '2021-10-10 ...' or integer like 150000000000. I think we need to process both of them too.

You are right. But when I set the timestamp type to int in data/csv's configuration , "integer like 150000000000" also shows errors like data type error. I guess the current td's "int" type may not support 150000000000's length.  Do you have some new idea?
Anyway, this PR has solved my problem.

sangshuduo:
> 
> 
> > thanks for your PR. The valid timestamp value can be either human-readable like '2021-10-10 ...' or integer like 150000000000. I think we need to process both of them too.
> 
> You are right. But when I set the timestamp type to int in data/csv's configuration , "integer like 150000000000" also shows errors like data type error. I guess the current td's "int" type may not support 150000000000's length. Do you have some new idea? Anyway, this PR has solved my problem.

it needs a 64bit integer actually. we use bigint in tdengine. C language use int64_t. some other software uses long.

sangshuduo:
> 
> 
> > thanks for your PR. The valid timestamp value can be either human-readable like '2021-10-10 ...' or integer like 150000000000. I think we need to process both of them too.
> 
> You are right. But when I set the timestamp type to int in data/csv's configuration , "integer like 150000000000" also shows errors like data type error. I guess the current td's "int" type may not support 150000000000's length. Do you have some new idea? Anyway, this PR has solved my problem.

Good to know you have a solution. Would you please improve your PR to support both forms?

Aries-Lee1991:
@t00350320 how is it going?

t00350320:
> @t00350320 how is it going?
en,I have finished.  please review :

For "The valid timestamp value can be either human-readable like '2021-10-10 ...' or integer like 150000000000", this pr will support both scenes. 

test scenes:
1、type = "timestamp" 
data/1.csv:
startTimestamp,endTimestamp,data
1605243000000,'2021-07-15 14:06:32.272',1
1605243000000,1605229140000,1

config/cfg.toml
fields = [
    { name = "startTimestamp",                  type = "timestamp" },
    { name = "endTimestamp",                  type = "timestamp" },
    { name = "data",            type = "int" },

select result as:
2020-11-13 12:35:00.000 | 2020-11-13 12:36:00.000
2020-11-13 12:50:00.000 | 2021-07-15 14:06:32.272

2、type="bigint"
data/1.csv:
startTimestamp,endTimestamp,data
1605243000000,1605229140000,1

config/cfg.toml
fields = [
    { name = "startTimestamp",                  type = "timestamp" },
    { name = "endTimestamp",                  type = "bigint" },
    { name = "data",            type = "int" },

select result as:
2020-11-13 08:58:00.000 |         1605229140000

sangshuduo (approved):
LGTM

PR #8636, opened 2021-11-10
[TS-593]<release>: change release script for ProDB

zyyang90:
[TS-593]<release>: change release script for ProDB


sangshuduo (approved):
LGTM

Issue #8340, opened 2021-10-20
Buffer not null terminated


wangmm0220:
https://github.com/taosdata/TDengine/blob/develop/src/tsdb/src/tsdbFile.c
Calling strncpy with a maximum size argument of 128 bytes on destination array pDFile->f.aname of size 128 bytes might leave the destination string unterminated.

![image](https://user-images.githubusercontent.com/6055324/138057460-1c3ad808-9487-43c8-a9bd-7b5e319191a7.png)

Hilbert-Yaa:
pick up challenge.

Aries-Lee1991:
> pick up challenge.

Assigned it to you.

Hilbert-Yaa:
@Aries-Lee1991 #8651

PR #8651, opened 2021-11-10
fix: #8340 __dest buffer is ensured null-terminated

hilbert-yaa:
fix the bug of __dest buffer after strncpy() not guaranteed null-terminated.


hzcheng (approved):

wangmm0220 (approved):
We can use tstrncpy function in this project. It is the same means as (*(TSDB_FILE_FULL_NAME(pMFile) + TSDB_FILENAME_LEN - 1) = '\0').

PR #8654, opened 2021-11-11
deal with tag prj function to support select csum(f1) from super-taba…

shenglian-zhou:
…le group by tbname limit n


hjxilinx (approved):
Approved.

PR #8661, opened 2021-11-11
taosdemo refactor split file

zhaoyanggh:


src/kit/taosdemo/src/demoFile.c
```diff
@@ -0,0 +1,1760 @@
+/*
```
sangshuduo: as opposite to demoCommand.c, how about demoJsonOpt.c for more explicit?

sangshuduo: demoCommand could be something like demoCommandOpt

sangshuduo (approved):
LGTM

afwerar (approved):
LGTM

PR #8672, opened 2021-11-12
[TS-612]update docs about quorum

liuyq-617:


boxiaobj:
please verify this scenario is working properly: replica 2, quorum 2, and salve vnode is offline, will it block the vgroup insert or not?

boxiaobj (approved):

liuyq-617:
> please verify this scenario is working properly: replica 2, quorum 2, and salve vnode is offline, will it block the vgroup insert or not?

yes, it will block insert

plum-lihui (approved):
ok

PR #8673, opened 2021-11-12
[TD-10873]<fix>: Fixed the problem that SMnodeMsg may still be used by dnodeMWriteQ thread after it is released by sdbWorker thread.

xiao-yu-wang:
[TD-10873]<fix>: Fixed the problem that SMnodeMsg may still be used by dnodeMWriteQ thread after it is released by sdbWorker thread.


dapan1121 (approved):

guanshengliang (approved):

guanshengliang:
There will be a memory leak if released before use, I accept and approve it

PR #8674, opened 2021-11-12
[TS-626]<fix>: fixed coredump

winshining:


DuanKuanJun (approved):

winshining:
Note: The current function can not sort the data in descending order, it will be refactored in the future.

PR #8702, opened 2021-11-16
[TS-618]<fix>: added one more check to differentiate normal and abnormal conditions

winshining:


stephenkgu (approved):
really sweet fix, :P

hjxilinx (approved):
I'm OK with this PR.

PR #8707, opened 2021-11-16
fix httpReadData will block one http thread until request finished

jianqingdu:
原来的代码效果和阻塞式socket一样，会在http处理线程里一直阻塞到读取一次HTTP请求的所有数据为止


src/plugins/http/src/httpServer.c
```diff
@@ -400,7 +400,7 @@ static bool httpReadData(HttpContext *pContext) {
     } else if (nread < 0) {
       if (errno == EINTR || errno == EAGAIN || errno == EWOULDBLOCK) {
         httpDebug("context:%p, fd:%d, read from socket error:%d, wait another event", pContext, pContext->fd, errno);
-        continue;  // later again
+        return false;
```
winshining, line 406: Excellent!

winshining (approved):
LGTM

src/plugins/http/src/httpServer.c
winshining: @jianqingdu I have discussed with @dapan1121 and we drew a conclusion: if the errno is **EINTR**, the read should immediately continue, while if the errno is EAGAIN or EWOULDBLOCK, return false.

jianqingdu: @winshining I agree with you, and make the change,  please recheck the PR

winshining (approved):
Good job!

dapan1121 (approved):

Issue #8690, opened 2021-11-16
Copy into fixed size buffer (STRING_OVERFLOW)


Aries-Lee1991:
https://github.com/taosdata/TDengine/blob/develop/src/client/src/tscServer.c

in line 1537

Copying p - > fullname without checking the length may overflow the 237 character fixed size string pcreate - > tablename.

![image](https://user-images.githubusercontent.com/48907085/141928610-4bcd1881-4543-4042-a08c-0eea4a2283cc.png)

xiaohu4313888:
pick up challenge

Aries-Lee1991:
> pick up challenge

Assigned it to you.

xiaohu4313888:
I have a problem, how do I solve this problem：
throw an exception when the size exceeds the preset size of 237 or
make the target pCreate save only the first 237 characters

hjxilinx:
Print this error message in the log file, and truncate the initial 237 characters of the full table name.

xiaohu4313888:
thanks i'll rewrite tonight

PR #8714, opened 2021-11-17
Fix/td 8690 fix size buffer (STRING_OVERFLOW)

xiaohu4313888:
I solved the problem by throwing an exception and retyping, and I'm not sure if the problem has been solved because I haven't had time to compile it yet.

problem can be described as Copying p - > fullname without checking the length may overflow the 237 character fixed size string pcreate - > tablename.
I added a module to check the length and re-enter it before copying,
I wonder if that solved the problem.

Resolves: #8690

Signed-off-by: Shen Zhi <m13120163046@163.com>

### DIFF
--- a/src/client/src/tscServer.c
+++ b/src/client/src/tscServer.c
@@ -1538,10 +1538,13 @@ int tscBuildCreateTableMsg(SSqlObj *pSql, SSqlInfo *pInfo) {
       //what p->fullname point is a char*
       //before the time we copy p->fullname to pCreate->tableName , we need to check the length of p->fullname
       if (strlen(p->fullname) > 237) {
-          tscError("failed to write size %d, which is over 237, just save the first 237 char here", strlen(p->fullname));
+        tscError("failed to write this name, which is over 237, just save the first 237 char here");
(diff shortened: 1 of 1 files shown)


xiaohu4313888:
If you have any questions, please leave a message and explain. I will modify it as required. Thank you

src/client/src/tscServer.c
```diff
@@ -1534,6 +1534,13 @@ int tscBuildCreateTableMsg(SSqlObj *pSql, SSqlInfo *pInfo) {
       pMsg += sizeof(SCreateTableMsg);
 
       SCreatedTableInfo* p = taosArrayGet(list, i);
+      //what pCreate->tableName point is a fixed char array which size is 237
+      //what p->fullname point is a char*
+      //before the time we copy p->fullname to pCreate->tableName , we need to check the length of p->fullname
+      while (strlen(p->fullname) > 237) {
+          printf("length of p->fullname should not exced 237\n");
```
hjxilinx: Print out the error message is not a good idea here, maybe use tscError  instead is a better idea?

hjxilinx (approved):
I'm OK with this PR.

PR #8729, opened 2021-11-18
Handle memory allocation failure when open vnode

haoyifan:
This patch fixes a few error handling of memory allocation
failure while opening vnodes.

bonus fix: also add a new error code to indicate failure
of opening vnode. Before the patch, the function would simply
return -1 when vnode open fails.


yihaoDeng (commented):
Please  define errno code  like 'TSDB_CODE_VND_ACTION_IN_PROGRESS',   Reference src file  terror.c & taoserror.h

yihaoDeng (approved):

guanshengliang (approved):

guanshengliang:
LGTM

PR #8739, opened 2021-11-19
[TS-453]<release>: kinghistorian release scripts

zyyang90:
[TS-453]<release>: kinghistorian release scripts


stephenkgu (approved):
A huge work done here, and is really error-prone this way, we need a refactor here to be less code-intrusive. Anyway I suggest let it in this time.

plum-lihui (approved):
LGMT

PR #8745, opened 2021-11-19
[TS-676]<fix>: fixed coredump when 'group by' was used

winshining:


stephenkgu (approved):
good fix.

hjxilinx (approved):
I'm OK with this PR.

PR #8751, opened 2021-11-20
fixed TS-426 on develop. group by can not ignore NULL

DuanKuanJun:


stephenkgu (approved):
same as #8751 , so I would approve & merge it.

PR #8752, opened 2021-11-20
[TD-10796]<fix>: when tag has escape char , can cause 'Syntax error in SQL', -2147483114

glzhao89:


yihaoDeng (commented):
no test case ?

glzhao89:
> 

This bug is actually reported long time back before the new rule that we request more develop selftest cases(actually some self-test cases are added in:
./general/parser/tagName_escape.sim
./general/parser/tbname_escape.sim
./general/parser/columnName_escape.sim
and merged into develop already but not covered this bug). 

And the full test cases are already added by test teams @happyguoxy but haven't submitted yet. We also use test teams cases to verify the fix for this bug. To avoid adding redundant test cases we'll use test teams test case for this feature this time. For future cases we will add more develop self test cases.

yihaoDeng:
> > 
> 
> This bug is actually reported long time back before the new rule that we request more develop selftest cases(actually some self-test cases are added in: ./general/parser/tagName_escape.sim ./general/parser/tbname_escape.sim ./general/parser/columnName_escape.sim and merged into develop already but not covered this bug).
> 
> And the full test cases are already added by test teams @happyguoxy but haven't submitted yet. We also use test teams cases to verify the fix for this bug. To avoid adding redundant test cases we'll use test teams test case for this feature this time. For future cases we will add more develop self test cases.

yihaoDeng:
LGTM

yihaoDeng (approved):

dapan1121 (approved):

PR #8770, opened 2021-11-23
[TD-11258]Unexpected result in "select expression from stable where ... interval() fill(null) group by" statement

dapan1121:


wangmm0220 (approved):
ok, remember add the test case

PR #8773, opened 2021-11-23
[TD-11258]Unexpected result in "select expression from stable where .…

dapan1121:
….. interval() fill(null) group by" statement


wangmm0220 (approved):
ok, remember add the test case

yihaoDeng (approved):
LGTM

PR #8776, opened 2021-11-23
[TD-10461]<test> add case for csum/mavg functions

wu-champion:


shenglian-zhou (approved):

xiaopingcs (approved):
Looks good, but it's nice to separate c code and python code in different PRs

PR #8777, opened 2021-11-23
[TS-527]<test>: add test case for state_window with subquery

xiaopingcs:
[TS-527]<test>: add test case for state_window with subquery


liuyq-617 (approved):
LGTM

plum-lihui (approved):
lgmt

PR #8779, opened 2021-11-24
[TS-707]<fix>: fixed crash when cleanup was called before creating all http threads

winshining:
If some errors occur when TD starts, cleanup will be called, but http threads may have not all been created yet.


dapan1121 (approved):

wangmm0220 (approved):
ok

PR #8788, opened 2021-11-25
[TD-11129]<test>: finish taosadapter perftest scripts

jiajingbin:
No need ci, main program is taosadapter_perftest.py which call other jmeter scripts.


xiaopingcs (approved):
Looks good

wu-champion (approved):
IP, port and similar parameters, it would be better if they can be turned into configuration options

jiajingbin:
> IP, port and similar parameters, it would be better if they can be turned into configuration options

I will optimize these in the next version, including deployment automation, monitoring automation, etc.

plum-lihui (approved):
LGMT

PR #8798, opened 2021-11-25
[TS-693]<release>: jh release script modify

zyyang90:
[TS-693]<release>: jh release script modify


stephenkgu (approved):
special release to jh

plum-lihui (approved):

PR #8799, opened 2021-11-25
Update the dependencies required for runing testcase

tomchon:


sangshuduo (approved):
LGTM

tomchon:
We need python3.8 or higher version to run test cases.Would we  mention it here? @sangshuduo

sangshuduo (changes requested):
we should explicitly require using python3.

sangshuduo:
> We need python3.8 or higher version to run test cases.Would we mention it here? @sangshuduo
 Yes.

liuyq-617:
uiautomation is needed on windows

plum-lihui (approved):
lgtm

PR #8804, opened 2021-11-26
[TD-11383]taos_fetch_block crashes during multithreaded operation on windows

dapan1121:


src/os/src/detail/osFile.c
```diff
@@ -43,8 +44,10 @@ void taosGetTmpfilePath(const char *fileNamePrefix, char *dstPath) {
     strcat(tmpPath, "-%d-%s");
   }
 
-  char rand[8] = {0};
-  taosRandStr(rand, tListLen(rand) - 1);
+  char rand[32] = {0};
+
+  sprintf(rand, "%" PRIu64, atomic_add_fetch_64(&seqId, 1));
+
   snprintf(dstPath, PATH_MAX, tmpPath, getpid(), rand);
```
xiao-yu-wang, line 51: The length of the 'rand' has increased from 8 to 32. Does the 'dstPath' also need to increase the length?

dapan1121: it should be OK, the max string length should be smaller than PATH_MAX.

glzhao89 (approved):

xiao-yu-wang (approved):

PR #8806, opened 2021-11-26
[TS-717]<test>: add test case for cachelast 

xiaopingcs:
[TS-717]<test>: add test case for cachelast


liuyq-617 (approved):
LGTM

wu-champion (approved):
suggest add it to function case

PR #8808, opened 2021-11-26
[TS-717]<test>: add test case for cachelast

xiaopingcs:
1. [TS-717]<test>: add test case for cachelast
2. fix random number issue: the schema defines a nchar(10) field, but random number -10000 will generate a string longer than 10 which will cause the case failure


liuyq-617 (approved):
LGTM

wu-champion (approved):
case is passed

PR #8814, opened 2021-11-26
[TS-705]<fix> tsBufNextPos vgroupIndex crash

DuanKuanJun:


stephenkgu (approved):
same as #8789, so I'll just merge it.

PR #8817, opened 2021-11-27
docs/TD-11198 Fix the format of the configuration parameter table

guanshengliang:


guanshengliang:
minor changes
and I will create a review task for wade

PR #8827, opened 2021-11-28
[TS-711]<fix>: windows shell limit 254 char error.

afwerar:


src/kit/shell/src/shellWindows.c
```diff
@@ -274,32 +276,34 @@ int32_t shellReadCommand(TAOS *con, char command[]) {
   // Read input.
   void *console = GetStdHandle(STD_INPUT_HANDLE);
   unsigned long read;
-  wchar_t c;
+  wchar_t c[SHELL_INPUT_MAX_COMMAND_SIZE];
```
sangshuduo: how about malloc()/free() instead of using an array on stack?

afwerar: ok

stephenkgu (approved):
good fix

src/kit/shell/src/shellWindows.c
```diff
@@ -274,32 +276,35 @@ int32_t shellReadCommand(TAOS *con, char command[]) {
...
+            free(cmd.command);
+            cmd.command = NULL;
+            free(c);
+            return 0;
+          } else {
+            shellPrintContinuePrompt();
+            updateBuffer(&cmd);
+          }
+          break;
+        case '\r':
```
sangshuduo, line 301: array 'c' be freed in case '\n' only, but seems the memory is not freed if the loop break from 'r'?

sangshuduo (approved):
LGTM

PR #8858, opened 2021-11-30
[TS-601]<fix>: no invalid database error in jdbc-restful with wrong db

zyyang90:
[TS-601]<fix>: no invalid database error in jdbc-restful with wrong db


stephenkgu (approved):
LGTM

xiaopingcs (approved):
looks good

PR #8883, opened 2021-12-01
[TD-11711]<fix>:rename method "executeSchemaless" to "insert"

huolibo:


sangshuduo (approved):
LGTM

dingbo8128 (approved):
approve name change. But I think the design can be refined.

sangshuduo:
> approve name change. But I think the design can be refined.

which component's design?

PR #8901, opened 2021-12-02
[TS-780]<fix> ts column is null crash

DuanKuanJun:


hjxilinx (approved):
I'm OK with this PR.

stephenkgu (approved):
LGTM

PR #8902, opened 2021-12-02
[ts-780]<fix> develop fix ts column is null

DuanKuanJun:


stephenkgu (approved):
same as #8901 for master

PR #8924, opened 2021-12-03
[TD-10522]<feature>:csharp support stmt

xleili:


src/connector/C#/src/TDengineDriver/TDengineDriver.cs
```diff
@@ -0,0 +1,392 @@
...
+        static extern public int AffectRows(IntPtr res);
+
+        [DllImport("taos", EntryPoint = "taos_field_count", CallingConvention = CallingConvention.Cdecl)]
+        static extern public int FieldCount(IntPtr res);
+
+        [DllImport("taos", EntryPoint = "taos_fetch_fields", CallingConvention = CallingConvention.Cdecl)]
+        static extern private IntPtr taos_fetch_fields(IntPtr res);
+        static public List<TDengineMeta> FetchFields(IntPtr res)
+        {
+            const int fieldSize = 68;
```
sangshuduo: why 68?

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
```diff
@@ -0,0 +1,392 @@
...
+            int fieldCount = FieldCount(res);
+            IntPtr fieldsPtr = taos_fetch_fields(res);
+
+            for (int i = 0; i < fieldCount; ++i)
+            {
+                int offset = i * fieldSize;
+
+                TDengineMeta meta = new TDengineMeta();
+                meta.name = Marshal.PtrToStringAnsi(fieldsPtr + offset);
+                meta.type = Marshal.ReadByte(fieldsPtr + offset + 65);
```
sangshuduo: why 65?

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
```diff
@@ -0,0 +1,392 @@
...
+                meta.size = Marshal.ReadInt16(fieldsPtr + offset + 66);
```
sangshuduo: why 66?

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
zitsen: To @sangshuduo :

```c
typedef struct taosField {
  char     name[65];
  uint8_t  type;
  int16_t  bytes;
} TAOS_FIELD;
```

68 is TAOS_FIELD struct size.
offset 65 is to get `type`.
offset 66 is to get `bytes`.

One quetion: should C# use the same definition of TAOS_FIELD or not? @xleili @sangshuduo

sangshuduo: I'm working to push the engine team to publish more definitions in the header file. Right now I make taosdef.h temporarily public. Could we refer to the definition from the header file instead of the magic number. 
At least, we can define them like, for example, FIELD_LENGTH 65, and FIELD_TYPE  (FIELD_NAME_LENGTH +1), etc.

zitsen: That will be much better to connectors. Adding constants for offsets is good for now.

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
xleili: read from offset 66, and read 2 bytes.

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
xleili: read form offest 65

src/connector/C#/src/TDengineDriver/TDengineDriver.cs
xleili: > 
I will define these number, and make the code easier to read. In next submit

zitsen (approved):
LGTM

sangshuduo (approved):
LGTM

PR #8934, opened 2021-12-06
[TD-10739]<test>: add test case for interp

wenzhouwww:
add test case for interp for TD-10736


happyguoxy:
Agree to add this special scenario

happyguoxy (approved):
Agree to add this special scenario

plum-lihui (approved):
lgtm

jiacy-jcy:
I think it is OK.

jiacy-jcy:
lgtm

PR #8935, opened 2021-12-06
[TD-10736]<test>:add interp test case

wenzhouwww:
add interp test case for TD-10736


wenzhouwww:
master  not support the latest interp function, the refactored interp function is in develop

PR #8936, opened 2021-12-06
[TS-780] add top buttom check

DuanKuanJun:


stephenkgu (approved):
same as #8393 for master

PR #8939, opened 2021-12-06
Hotfix/sangshuduo/td 11842 split taostools targz

sangshuduo:


tomchon (approved):
are we going to package taosdump  and taosdemo independently?

sangshuduo:
> are we going to package taosdump and taosdemo independently?

yes, taos-tools will include taosdump and taosBenchmark(taosdemo) as a separated package.

plum-lihui (approved):
lgtm

PR #8946, opened 2021-12-06
[TS-816]<fix>: query cannot stop in windows

zyyang90:
[TS-816]<fix>: query cannot stop in windows


stephenkgu (approved):
same as #8845 for master

hjxilinx (approved):
I'm OK with this PR.

PR #8956, opened 2021-12-07
[TS-832]<feature>:check and output drop tables by specific errNo

kailixu:


stephenkgu (approved):
need this patch tonight.

PR #8969, opened 2021-12-08
[TS-787]<enhance>:release lock during write socket

kailixu:


stephenkgu (approved):
same as #8968 for master

PR #8974, opened 2021-12-08
[TD-11711]<fix>:make schemaless insert as an independent class, not associated with statement

huolibo:


dingbo8128 (approved):
Approve. But it will be better to add the followin override method:
```
 public void write(ArrayList<String> lines, .....);
```

zyyang-taosdata (approved):
LGTM

PR #8995, opened 2021-12-08
[TS-854]<fix>: add forceDataRow option in taos.cfg

kailixu:


kailixu:
close as to utilize final solution.

PR #9010, opened 2021-12-09
Test/td 10700 d

happyguoxy:
[TD-10700]<test>The escape char "`" can be used for both tag name and column name


tomchon (approved):
i suggest that the path be made into a parameterized form
--------------
os.system("rm -rf 0-management/1-stable/create_col_tag.py.sql")

plum-lihui (approved):
lgtm

tomchon (approved):
LGTM

PR #9013, opened 2021-12-09
[TS-854]<fix>:generate final row from origin

kailixu:
[TS-854]<fix>:generate final row from origin


hzcheng (approved):

hjxilinx (approved):
I'm OK with this PR.

PR #9044, opened 2021-12-11
[TS-836]<feature>: getLong return a timestamp with good precision

zyyang90:
[TS-836]<feature>: getLong return a timestamp with good precision


stephenkgu (approved):
LGTM

huolibo (commented):
If there is no time column in the result , the time precision still occur in restful api？

huolibo (approved):

dingbo8128 (commented):
LGTM

PR #9085, opened 2021-12-14
fix error if json value is negative float

wangmm0220:


glzhao89 (approved):

shenglian-zhou (approved):
better ways to find whether it is double or big int?

glzhao89:
> better ways to find whether it is double or big int?
There're two util functions:
bool isValidInteger(char *str)
bool isValidFloat(char *str)
can be used to check if a string is integer/float.
But in this case the input is double already, unless use cJSON structure's numberstring.

PR #9088, opened 2021-12-14
[TD-11980]<docs>: refine taosdemo docs

sangshuduo:


wu-champion (changes requested):
the changes looks OK，but the parameter description part of the "Use taosdemo for query and subscription testing" paragraph shows confusion

wu-champion (approved):
LGTM

PR #9090, opened 2021-12-14
TD-11943: fix bug core dump related to select tbname, max(col)+val from childtable

shenglian-zhou:


wangmm0220 (approved):

hjxilinx (approved):
I'm OK with this PR.

PR #9140, opened 2021-12-16
Fix/td 11789 If client and server charset are not the same, error will be appear in where string condition

wangmm0220:


shenglian-zhou (approved):
test?

src/client/src/tscSQLParser.c
```diff
@@ -5034,6 +5040,21 @@ int32_t handleNeOptr(tSqlExpr** rexpr, tSqlExpr* expr) {
   return TSDB_CODE_SUCCESS;
 }
 
+static void convertWhereStringCharset(tSqlExpr* pRight){
+  if(pRight->value.nType != TSDB_DATA_TYPE_BINARY){
+    return;
+  }
+
+  char newData[TSDB_MAX_NCHAR_LEN] = {0};
```
shenglian-zhou: big allocation on stack. the max length is pRight->value.nLen, malloc it then free?

src/client/src/tscSQLParser.c
```diff
@@ -5034,6 +5040,21 @@ int32_t handleNeOptr(tSqlExpr** rexpr, tSqlExpr* expr) {
...
+  int len = 0;
+  if(!taosMbsToUcs4(pRight->value.pz, pRight->value.nLen, newData, TSDB_MAX_NCHAR_LEN, &len)){
+    tscError("nchar in where condition mbsToUcs4 error");
+  }
+  pRight->value.pz = realloc(pRight->value.pz, len);
```
glzhao89: better check return value of realloc. If allocate failure, the original pointer will be overwritten by NULL thus cause memory leak.

glzhao89 (approved):

wangmm0220:
> test?

test report:https://jira.taosdata.com:18090/display/DEV/TD-11789

Because it is need to change the charset of configure, it is hard add test case in python test case. I just test it in my machine.

PR #9155, opened 2021-12-17
[TD-12223]<fix>:fix jdbc fetch time before 1970

huolibo:


src/connector/jdbc/src/main/java/com/taosdata/jdbc/TSDBResultSetRowData.java
```diff
@@ -443,16 +443,19 @@ public void setTimestamp(int col, long ts, int precision) {
...
             case 1: {
                 milliseconds = ts / 1_000;
                 fracNanoseconds = (int) (ts * 1_000 % 1_000_000_000);
+                fracNanoseconds = fracNanoseconds < 0 ? 1_000_000 + fracNanoseconds : fracNanoseconds;
                 break;
             }
             case 2: {
                 milliseconds = ts / 1_000_000;
                 fracNanoseconds = (int) (ts % 1_000_000_000);
+                fracNanoseconds = fracNanoseconds < 0 ? 1_000 + fracNanoseconds : fracNanoseconds;
```
dingbo8128: I think milliseconds should also be adjusted in case fracNanoseconds < 0

dingbo8128 (changes requested):

PR #9218, opened 2021-12-20
[TS-932]<test>: add test case for client clientMerge option

xiaopingcs:
[TS-932]<test>: add test case for client clientMerge option


liuyq-617 (approved):
LGTM

plum-lihui (approved):

PR #9220, opened 2021-12-20
[TD-12344]<test> this is case to generate draft for taoshell crash about session function

wenzhouwww:
this is case to generate draft for taoshell crash about session function!


happyguoxy (approved):
LGTM

plum-lihui (approved):
lgtm

PR #9234, opened 2021-12-21
[TD-12204]<test>core:qExtbuffer.c:1193

happyguoxy:


wenzhouwww (approved):
LGTM ，I think If it is for a specific bug, submit the smallest unit of use case that triggers the case.

PR #9271, opened 2021-12-22
[TD-11970] <tets> add case for TD-11970

wu-champion:


glzhao89 (approved):

xiaopingcs (approved):
Looks good

wu-champion:
frame is change， rePR

PR #9283, opened 2021-12-22
[TD-12229]<test>: add test case for error result about union all

wenzhouwww:
add test case for error result about union all when between stable


xiaopingcs (changes requested):
Please fix conflicts.

wenzhouwww:
> Please fix conflicts.

when other PR merge into develop , I will solve conficts  and rerun  ASP . because it will change file fulltest_query.sh

PR #9284, opened 2021-12-22
[TD-12165]<test>: add test case for alis _c0

wenzhouwww:
add test case for alis _c0


wenzhouwww:
include into PR https://github.com/taosdata/TDengine/pull/9293

PR #9301, opened 2021-12-23
[TD-12195]<fix>(insert): adjust the wal overhead to 1/3 so that wal limit is not exceeded

shenglian-zhou:


glzhao89 (approved):

src/client/src/tscParseLineProtocol.c
```diff
@@ -1061,7 +1061,7 @@ static int32_t insertChildTablePointsBatch(TAOS* taos, char* cTableName, char* s
   tscDebug("SML:0x%"PRIx64" insert child table table %s of super table %s : %s", info->id, cTableName, sTableName, sql);
 
   size_t rows = taosArrayGetSize(rowsBind);
-  size_t maxBatchSize = TSDB_MAX_WAL_SIZE/rowSize * 4 / 5;
+  size_t maxBatchSize = TSDB_MAX_WAL_SIZE/rowSize * 2 / 3;
```
wangmm0220, line 1063: it will be better if it can be changed by dynamic. Otherwise it may be happen again if input data change.

wangmm0220 (approved):

PR #9312, opened 2021-12-23
Return accurate string to Lua.

robotspace:
Use lua_pushlstring instead of lua_pushstring to return accurate string to Lua.


sangshuduo (approved):
LGTM

PR #9345, opened 2021-12-23
Feature/3.0 mhli

castermode:
TD-12484 simulate a raft kv store, for self testing


castermode:
TD-12484 simulate a raft kv store, for self testing

guanshengliang (approved):

PR #9415, opened 2021-12-27
[TS-995]<fix>(query): validate pTableCheckInfo when querying last/las…

kailixu:


stephenkgu (approved):

stephenkgu:
same as #9417 for develop

PR #9420, opened 2021-12-27
[TS-1004]<enhance>(insert):disable warning for batch insert data with…

kailixu:
… identical TS


kailixu:
Should set disordered in order to trigger sorting and merging identicalTS in later steps.

PR #9428, opened 2021-12-28
TD-12275: fix core dump caused by num of rows multiple by column bytes exceeds INT32_MAX

shenglian-zhou:


wangmm0220 (approved):

src/query/src/qExecutor.c
```diff
@@ -3586,31 +3586,34 @@ void setDefaultOutputBuf(SQueryRuntimeEnv *pRuntimeEnv, SOptrBasicInfo *pInfo, i
...
+      char* p = realloc(pColInfo->pData, ((size_t)newSize) * pColInfo->info.bytes);
       if (p != NULL) {
         pColInfo->pData = p;
 
         // it starts from the tail of the previously generated results.
         pBInfo->pCtx[i].pOutput = pColInfo->pData;
         (*bufCapacity) = newSize;
       } else {
-        // longjmp
+        size_t allocateSize = ((size_t)(newSize)) * pColInfo->info.bytes;
```
glzhao89, line 3605: allocatedSize can be declared before the if condition and used in realloc also

glzhao89 (approved):

PR #9483, opened 2021-12-29
[TS-995]<fix>(query): validate pTableCheckInfo when querying last/last_row

kailixu:


kailixu:
should use 2.4 instead of 2.4.0.0

PR #9498, opened 2021-12-30
[TD-11999]<fix>(other): remove double quotes from restful_info table using underscore instead

glzhao89:


src/plugins/monitor/src/monMain.c
```diff
@@ -463,7 +463,7 @@ static void monBuildMonitorSql(char *sql, int32_t cmd) {
     int pos = snprintf(sql, SQL_LENGTH,
                        "create table if not exists %s.restful_info(ts timestamp", tsMonitorDbName);
     for (int i = 0; i < tListLen(monHttpStatusTable); ++i) {
-      pos += snprintf(sql + pos, SQL_LENGTH, ", `%s(%d)` int",
+      pos += snprintf(sql + pos, SQL_LENGTH, ", %s_%d int",
```
shenglian-zhou: change the second parameter of snprintf to free space size, SQL_LENTH_POS?

glzhao89: Ok, will change this to SQL_LENGTH - usedBytes to prevent potential buffer overflow.

PR #9500, opened 2021-12-30
[TS-802]<bug>(query): develop fixed first last not support desc order

DuanKuanJun:


DuanKuanJun:
replace with TS-802-D1 branch, so close this

PR #9505, opened 2021-12-30
qScript.c: potential buffer overrun

freemine:


freemine:
seems like there's typo in test case, which results in CI failure.
```
tests/pytest/tools/taosdemoAllTest/taosdemoTestInsertWithJson.py:151
tests/pytest/tools/taosdemoAllTest/insert-offset.json:46
```

yihaoDeng (approved):

freemine:
[see here: https://github.com/taosdata/TDengine/pull/9728](https://github.com/taosdata/TDengine/pull/9728)
#9728

PR #9567, opened 2022-01-04
[TS-927]<fix>(query): fix fill(linear) with null values

stephenkgu:


stephenkgu:
same as #9570 for develop.

guanshengliang (approved):

PR #9657, opened 2022-01-06
[TS-925]<fix>(query): fixed inconsistant primary col name

winshining:


stephenkgu:
same as #9658 for v2.4

PR #9666, opened 2022-01-07
[TD-12723]<feature>(query): add taos_result_block api for taosAdaptor ulimited pulling function

glzhao89:


huskar-t (commented):
Need to add the api to the taos.def file

glzhao89:
> 

added api in taos.def file

huskar-t (approved):
LGTM

src/client/src/tscSql.c
```diff
@@ -547,6 +547,28 @@ int taos_fetch_block(TAOS_RES *res, TAOS_ROW *rows) {
...
+  SSqlRes *pRes = &pSql->res;
+
+  if (pCmd == NULL ||
+      pRes == NULL ||
+      pRes->qId == 0 ||
+      pRes->code == TSDB_CODE_TSC_QUERY_CANCELLED ||
+      pCmd->command == TSDB_SQL_RETRIEVE_EMPTY_RESULT ||
+      pCmd->command == TSDB_SQL_INSERT) {
+    return NULL;
+  }
```
shenglian-zhou, line 567: when the fetch is initiated so that the result is filled?

PR #9815, opened 2022-01-14
jdbc: make jdbc-target a non-always-run-rule

freemine:
It's annoying that jdbc-target is always built even if there's no change in source since last successful build.
With this patch, built-time is also saved.
Besides, also add both ```taos``` and ```taos-static``` as it's dependent target.


sangshuduo (approved):
LGTM

zyyang-taosdata (approved):
LGTM

PR #10144, opened 2022-02-08
update stb

guanshengliang:


cApTO23 (commented):
so if '[primary TAB IDs]'+ <stdin>= val A+; deriv.IDs grouped w/prim. must be = to |A| or the offset/thereof. returning none‽

PR #10185, opened 2022-02-10
add comment for stb

guanshengliang:


cApTO23 (commented):
I inferred the wrong implementation(s) of the guidelines in my last comment.  I  strugling to keep up.  How do I correct my previous commits without a complete void? With respect to DEF, sig-dig IDs, mobile limitations using CLI
REF:PR 10185

sangshuduo:
> I inferred the wrong implementation(s) of the guidelines in my last comment. I strugling to keep up. How do I correct my previous commits without a complete void? With respect to DEF, sig-dig IDs, mobile limitations using CLI REF:PR 10185

May I know what is your suggestion about our PR or code? Could you please comment inline?

PR #10196, opened 2022-02-10
[TD-13408]<test>: move examples back to TDegnine

sangshuduo:
for 3.0


bding2018:
MULTIBIND is not supported.

sangshuduo:
> MULTIBIND is not supported.

Could you please explain what issue you are facing?

PR #10221, opened 2022-02-11
[TD-11946]<feat>: implement jdbc with websocket and get/parse JSON resultSet 

huolibo:


dingbo8128 (approved):
LGTM

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
```diff
@@ -0,0 +1,593 @@
...
+            case TSDBConstants.TSDB_DATA_TYPE_FLOAT:
+                return row.getFloat(colIndex);
+            case TSDBConstants.TSDB_DATA_TYPE_DOUBLE:
+                return row.getDouble(colIndex);
+            case TSDBConstants.TSDB_DATA_TYPE_TIMESTAMP:
+                return parseTimestampColumnData(row, colIndex);
+            case TSDBConstants.TSDB_DATA_TYPE_BINARY:
+                return row.getString(colIndex) == null ? null : row.getString(colIndex).getBytes();
+            case TSDBConstants.TSDB_DATA_TYPE_NCHAR:
+                return row.getString(colIndex) == null ? null : row.getString(colIndex);
```
sangshuduo: seems no need ternary operator?

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
```diff
@@ -0,0 +1,593 @@
...
+                    long nanoAdjustment = value % 1000_000_000L;
+                    return Timestamp.from(Instant.ofEpochSecond(epochSec, nanoAdjustment));
+                }
+            }
+            case UTC: {
+                String value = row.getString(colIndex);
+                if (value.lastIndexOf(":") > 19) {
+                    ZonedDateTime parse = ZonedDateTime.parse(value, rfc3339Parser);
+                    long nanoAdjustment;
+                    if (value.length() > 32) {
```
sangshuduo: is there a formal way to pass precision instead of check the value length?

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
```diff
@@ -0,0 +1,593 @@
...
+                        nanoAdjustment = fractionalSec * 1000_000L;
+                        this.timestampPrecision = TimestampPrecision.MS;
+                    }
+                    ZoneOffset zoneOffset = ZoneOffset.of(value.substring(value.length() - 5));
+                    Instant instant = Instant.ofEpochSecond(epochSec, nanoAdjustment).atOffset(zoneOffset).toInstant();
+                    return Timestamp.from(instant);
+                }
+            }
+            case STRING:
+            default: {
```
sangshuduo: indent

sangshuduo (commented):
see my comments

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
huolibo: ok, remove the redundant code.

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
huolibo: There are many representations of UTC time, like `yyyy-MM-ddTHH:mm:ss.SSSSSSSSS+0x:00`,  `yyyy-MM-ddTHH:mm:ss.SSS+0x:00`, `yyyy-MM-ddTHH:mm:ss.SSSSSSSSS+0x00`, `yyyy-MM-ddTHH:mm:ss.SSS+0x00` and so on.  I need to deal with them separately. Do other languages have a more efficient way of doing this？

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
huolibo: This is OK. This means：If it's a STRING, the processing is the same as default

src/connector/jdbc/src/main/java/com/taosdata/jdbc/ws/JSONResultSet.java
sangshuduo: we have the API taos_result_precision(TAOS_RES *res) to return precision. Can we get it from remote first?

dingbo8128 (approved):

src/connector/jdbc/src/main/java/com/taosdata/jdbc/rs/RestfulDriver.java
```diff
@@ -77,26 +78,30 @@ public Connection connect(String url, Properties info) throws SQLException {
...
+                if (!client.isAuth()) {
                     throw new SQLException("auth failure");
                 }
             } catch (URISyntaxException e) {
                 throw new SQLException("Websocket url parse error: " + loginUrl, e);
             } catch (InterruptedException e) {
                 throw new SQLException("creat websocket connection has been Interrupted ", e);
             }
-            return new WSConnection(url, props, transport, database, true);
+            // TODO fetch Type from config
```
sangshuduo, line 102: still need to do it in the future?

sangshuduo (approved):
LGTM but not sure if we should leave a TODO

PR #10269, opened 2022-02-15
[TD-13300]<fix>: docker stop takes 10 seconds

sangshuduo:


packaging/docker/dockerbuild.sh
```diff
@@ -89,7 +89,7 @@ cp -f ${comunityArchiveDir}/${pkgFile}  .
 echo "dirName=${dirName}"
 
 
-docker build --rm -f "Dockerfile"  --network=host -t tdengine/tdengine-${dockername}:${version} "." --build-arg pkgFile=${pkgFile} --build-arg dirName=${dirName}
+docker build --rm -f "Dockerfile"  --network=host -t tdengine/tdengine-${dockername}:${version} "." --build-arg pkgFile=${pkgFile} --build-arg dirName=${dirName} --build-arg cpuType=${cpuType}
```
zitsen, line 92: LGTM, but the `cpuType` here we use currently is `x64` not `amd64`, which will cause 404 page error while downloading.

zitsen, line 92: LGTM, but the cpuType here we use currently is x64 not amd64, which will cause 404 page error while downloading.

zitsen (commented):
see my comment

packaging/docker/dockerbuild.sh
sangshuduo: not really. the cpuType will be re-specified to amd64 if use x64 in release script now. I verified tini-amd64 will be downloaded as expected.

zitsen (approved):
LGTM

packaging/docker/dockerbuild.sh
zitsen: That will be ok.

Eavanyuan (approved):
approve

PR #10353, opened 2022-02-22
[TD-11946]<feature>: parse block data and handle null value

huolibo:


src/connector/jdbc/src/main/java/com/taosdata/jdbc/utils/NullType.java
```diff
@@ -11,81 +11,97 @@ public String toString() {
...
     }
-    
+
     public static boolean isTinyIntNull(byte val) {
         return val == Byte.MIN_VALUE;
     }
-    
+
+    public static boolean isUnsignedTinyIntNull(byte val) {
+        return val == (byte) 0xFF;
```
sangshuduo, line 20: we should use taos_is_null() from now.

src/connector/jdbc/src/main/java/com/taosdata/jdbc/utils/NullType.java
```diff
@@ -11,81 +11,97 @@ public String toString() {
...
+    }
+
     public static boolean isSmallIntNull(short val) {
         return val == Short.MIN_VALUE;
     }
 
+    public static boolean isUnsignedSmallIntNull(short val) {
+        return val == (short) 0xFFFF;
```
sangshuduo, line 28: same concern

sangshuduo (changes requested):
see my comments

dingbo8128 (approved):
LGTM

sangshuduo (approved):
LGTM

PR #10397, opened 2022-02-24
A few changes as outlined below

cdiwadkar16:
1. In the first sentence "TDengine includes server, client, and ecological software ..." I propose changing the word "ecological" to "ecosystem"
2. Change "After 2.4" to "Starting with 2.4"
3. Change "including RESTful service" to "and includes a RESTful service"
4. Change "early version" to "earlier version"
5. Change "For the time being, using Docker to deploy the client or server of TDengine for production environments is not recommended." to "For the time being, we do not recommend using Docker to deploy the TDengine server or client in production environments."
6. Change "TDengine is very easy to install, from download to successful installation in just a few seconds." to "TDengine is very easy to run; download to successful installation takes just a few seconds."
7. Split the second sentence in the Install from Package section into 2 sentences with a period instead of ";"
8. Under Quick Launch, change "Then check if the service is working now." to "Then check if the system is working."
9. Under TDengine Shell Command Line change "Besides the SQL commands" to "Besides executing SQL commands"

### DIFF
--- a/documentation20/en/02.getting-started/docs.md
+++ b/documentation20/en/02.getting-started/docs.md
@@ -2,7 +2,7 @@
 
 ## <a class="anchor" id="install"></a>Quick Install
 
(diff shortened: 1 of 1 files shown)


jtao1735 (approved):

sangshuduo (approved):
LGTM

PR #10437, opened 2022-02-28
change the big table for configuration to list

dingbo8128:


chiyitao (commented):
Agree with the modification.

PR #10538, opened 2022-03-04
[TD-13785]<fix>: change method for jdbc judge update or query

huolibo:


dingbo8128 (commented):
Why not use the same way as JNI ?

huolibo:
> Why not use the same way as JNI ?

because  restful API can not call method in taos.lib

dingbo8128 (approved):
LGTM

zyyang-taosdata (approved):
LGTM

PR #10628, opened 2022-03-08
[TS-1289][TS-1255][TS-1286]: add test cases

xiaopingcs:
[TS-1289][TS-1255][TS-1286]: add test cases

### DIFF
--- a/tests/pytest/tools/taosdumpTest2.py
+++ b/tests/pytest/tools/taosdumpTest2.py
@@ -100,7 +100,7 @@ class TDTestCase:
         tdSql.checkData(0, 0, 'st')
 
         tdSql.query("select count(*) from t1")
-        tdSql.checkData(0, 0, self.numberOfRecords)
+        tdSql.checkData(0, 0, self.numberOfRecords)        
 
         # test case for TS-1225
         tdSql.execute("create database test")
@@ -126,6 +126,7 @@ class TDTestCase:
 
         tdSql.query("select * from stb")
         tdSql.checkRows(1)
+        os.system("rm -rf dump_result.txt")
 
     def stop(self):
         tdSql.close()


tests/pytest/tools/taosdumpTest2.py
```diff
@@ -91,6 +100,30 @@ def run(self):
         tdSql.query("select count(*) from t1")
         tdSql.checkData(0, 0, self.numberOfRecords)
 
+        tdSql.execute("create database test")
+        tdSql.execute("use test")
+        tdSql.execute("create table stb(ts timestamp, c1 binary(16000), c2 binary(16000), c3 binary(10000)) tags(t1 nchar(256))")
```
sangshuduo: how about using a max boundary value to test the max length row?

tests/pytest/tools/taosdumpTest2.py
```diff
@@ -91,6 +100,30 @@ def run(self):
...
+        tdSql.execute("drop database test")
+        tdSql.query("show databases")
+        tdSql.checkRows(0)
+
+        os.system("%staosdump -i /tmp" % binPath)
+
+        tdSql.execute("use test")
+        tdSql.error("show vnodes '' ")
+        tdSql.query("show stables")
+        tdSql.checkRows(1)
```
sangshuduo, line 124: it would be better if we check the value with the old one to avoid it being trimmed.

sangshuduo (approved):
LGTM with suggestions.

tests/pytest/tools/taosdumpTest2.py
xiaopingcs: the value is also checked in line 122

tests/pytest/tools/taosdumpTest2.py
xiaopingcs: updated

PR #10647, opened 2022-03-09
<fix> : fix crash_gen support restful mode by deploy taosadapter

wenzhouwww:
because of resultful now is integrated in taosAdapter ,so extend taosadapter deployment for every TDengine instance .

by the way , taosadapter and TDengine are stateless, so start taosAdapters are separately, not integrated into TDengine instance state management.

### DIFF
--- a/tests/pytest/crash_gen/crash_gen_main.py
+++ b/tests/pytest/crash_gen/crash_gen_main.py
@@ -1701,7 +1701,7 @@ class TdSuperTable:
             dbc.execute("DROP TaBLE {}".format(fullTableName))
         else:
             if not skipCheck:
-                raise CrashGenError("Cannot drop non-existant super table: {}".format(fullTableName))
+                raise CrashGenError("Cannot drop non-existant super table: {}".format(self._stName))
 
     def exists(self, dbc):
         dbc.execute("USE " + self._dbName)


stevel032 (approved):
Looks good to me, please go ahead.

I do have a quick suggestion though, looking at the introduction of new member variable "self._fullTableName = dbName + '.' + stName". This new variable has duplicate/redundant information from other variables. This creates potential problem of inconsistency (what if one variable changes but the other does not). So I would suggest that we simply implement a getFullTableName() method to dynamically construct the table name. We sacrifice a very small amount of performance to get a small improvement in reliability.

wenzhouwww:
> Looks good to me, please go ahead.
> 
> I do have a quick suggestion though, looking at the introduction of new member variable "self._fullTableName = dbName + '.' + stName". This new variable has duplicate/redundant information from other variables. This creates potential problem of inconsistency (what if one variable changes but the other does not). So I would suggest that we simply implement a getFullTableName() method to dynamically construct the table name. We sacrifice a very small amount of performance to get a small improvement in reliability.

Good suggestion ！
it  can avoid the problem of changing the table name  in processing ,   I will add an attribute method and dynamically obtain the current value of the fullTableName when using it.

PR #10731, opened 2022-03-14
[TD-13820]<feature>(taosAdapter): compile taosadapter debug version

huskar-t:


sangshuduo (approved):
LGTM. however why not just combine the different parts into one section.

src/plugins/CMakeLists.txt
```diff
@@ -67,6 +67,19 @@ ELSE ()
           COMMAND cmake -E copy ./example/config/taosadapter.toml ${CMAKE_BINARY_DIR}/test/cfg/
           COMMAND cmake -E copy ./taosadapter.service ${CMAKE_BINARY_DIR}/test/cfg/
           )
+      ExternalProject_Add(taosadapter-debug
+          PREFIX        "taosadapter-debug"
```
zitsen: LGTM. But seems no difference with the next change block (ignore the os cmake condition).

zitsen (approved):

sangshuduo (approved):
LGTM

PR #10812, opened 2022-03-18
add taosbenchmark sample json and csv files

zhaoyanggh:


sangshuduo (changes requested):
prefer to manage example files in the taos-tools project, just keep minimal change in TDengine if need. And please follow the rule of branch naming.

PR #10813, opened 2022-03-18
add taosbenchmark sample json and csv files

zhaoyanggh:


sangshuduo (changes requested):
should use a https://github.com/taosdata/TDengine/pull/10850 alike PR instead of this one

PR #10819, opened 2022-03-18
Test/xiaolei/td 13708 test csharp subscribe

xleili:


src/connector/C#/src/test/FunctionTest/SubscribeAsync.cs
```diff
@@ -0,0 +1,241 @@
...
+                        for (int i = 0; i < actualResData.Count; i++)
+                        {
+                            // output.WriteLine("index:{0},expectResData:{1},actualResData:{2}", i, expectResData[i], actualResData[i]);
+                            Assert.Equal(expectResData2[i].ToString(), actualResData[i]);
+                        }
+                    }
+                }
+                else
+                {
+                    output.WriteLine($"async query data failed, failed code {code}");
```
sangshuduo: how about "error code" here? and can we provide more info provided by taos_errstr() interface?

sangshuduo (commented):
see my comments

src/connector/C#/src/test/FunctionTest/SubscribeAsync.cs
xleili: fixed

src/connector/C#/src/test/FunctionTest/SubscribeAsync.cs
```diff
@@ -0,0 +1,257 @@
...
+                            // output.WriteLine("index:{0},expectResData:{1},actualResData:{2}", i, expectResData[i], actualResData[i]);
+                            Assert.Equal(expectResData[i].ToString(), actualResData[i]);
+                        }
+                    }
+
+
+                }
+                else if (taosRes != IntPtr.Zero)
+                {
+                    output.WriteLine($"async query data failed, failed code {TDengine.Error(taosRes)}");
```
sangshuduo: should be: error code xx, reason: xxxxx

sangshuduo (changes requested):
see my comments

zitsen (approved):
LGTM

sangshuduo (approved):
LGTM

PR #10887, opened 2022-03-22
[TD-14218]<fix>: remove unused variables

huolibo:


sangshuduo:
workaround it by cmake is better to those who does not care Mac for now

PR #11057, opened 2022-03-28
[TS-1363]<fix>: Mutex is not initialized

54liuyao:


stephenkgu (approved):

stephenkgu:
same as #11053 for v2.4

PR #11246, opened 2022-04-05
add node.js sample codes

dingbo8128:


sangshuduo (approved):
LGTM for most parts. still need to give a reasonable and verified version range

PR #11303, opened 2022-04-07
TD-14565:log function with default param base e

shenglian-zhou:


wangmm0220 (approved):

src/common/src/texpr.c
```diff
@@ -2041,13 +2061,20 @@ void vectorMathFunc(int16_t functionId, tExprOperandInfo *pInputs, int32_t numIn
     if (!hasNullInputs) {
       switch (functionId) {
         case TSDB_FUNC_SCALAR_LOG: {
-          assert(numInputs == 2);
-          double base = 0;
-          GET_TYPED_DATA(base, double, pInputs[1].type, inputData[1]);
+          assert(numInputs == 1 || numInputs == 2);
```
gccgdb1234: I guess this line is useless, because the error has already been caught at line 1202

gccgdb1234: by the way, we should NOT use assert() in product code. There have been already many asserts, let's at least not add new ones.

src/common/src/texpr.c
```diff
@@ -2041,13 +2061,20 @@ void vectorMathFunc(int16_t functionId, tExprOperandInfo *pInputs, int32_t numIn
...
+          double base = M_E;
+          if (numInputs == 2) {
+            GET_TYPED_DATA(base, double, pInputs[1].type, inputData[1]);
+          }
           double v1 = 0;
           GET_TYPED_DATA(v1, double, pInputs[0].type, inputData[0]);
-          double result = log(v1) / log(base);
-          SET_TYPED_DATA(outputData, pOutput->type, result);
+          if (numInputs == 2) {
```
gccgdb1234, line 2070: if is not necessary here, because base has been initialized regardless of numInputs=1 or 2, here we just set result=log(v1)/log(base). In case of numInputs=1, log(base) is exactly 1, right?

glzhao89 (approved):

src/common/src/texpr.c
shenglian-zhou: Improve precision. log(M_E) is not exactly 1.

src/common/src/texpr.c
shenglian-zhou: yes. I will remove it.

src/common/src/texpr.c
shenglian-zhou: save two time-consuming computations, log(M_E) and double division when numInput is 1.

gccgdb1234: ok, accepted

PR #11405, opened 2022-04-12
fix: resolve win test error.

afwerar:


sangshuduo (changes requested):
taosAdapter need --monitor.writeToTD=false to avoid creating log db. we should investigate how pass parameter to taosadapter for mitty or use other way to make taosadapter not to create log db.

gccgdb1234 (approved):

sangshuduo (approved):
LGTM. better to prompt user "taosadapter not found" if it's empty.

PR #11413, opened 2022-04-12
feat(query): add cast function

glzhao89:
TD-14242

### DIFF
--- a/include/common/ttypes.h
+++ b/include/common/ttypes.h
@@ -141,6 +141,43 @@ typedef struct {
     }                                          \
   } while (0)
 
+#define NUM_TO_STRING(_inputType, _input, _outputBytes, _output)     \
+  do {                                         \
+    switch (_inputType) {                      \
+      case TSDB_DATA_TYPE_TINYINT:             \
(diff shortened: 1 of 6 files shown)


include/common/ttypes.h
```diff
@@ -141,6 +141,43 @@ typedef struct {
     }                                          \
   } while (0)
 
+#define NUM_TO_STRING(_inputType, _input, _outputBytes, _output)     \
```
gccgdb1234, line 144: why not inline function? It's hard to maintain such a big "macro"

glzhao89: Agreed big macro is hard to maintain. This macro is simply ported from 2.6 so pretty much following the same convention if look at other macros defined above this code block: e.g. SET_TYPED_DATA/GET_TYPED_DATA etc. 
Not sure if inlining this as function would be cost efficient as this code block is pretty large and contains many conditional branches(assume there's penalty for cpu to do branch prediction).

gccgdb1234: ok, if it's ported code, let's keep it.

PR #11423, opened 2022-04-12
feat(query): group by multi columns

stephenkgu:


stephenkgu:
same as #11315 for v2.4

PR #11466, opened 2022-04-13
feat(tsdb): delete time series data

DuanKuanJun:


src/client/src/tscSql.c
```diff
@@ -416,6 +416,13 @@ int taos_affected_rows(TAOS_RES *tres) {
   return pSql->res.numOfRows;
 }
 
+int taos_affected_tables(TAOS_RES *tres) {
```
hjxilinx, line 419: why is this API needed?

src/util/inc/tqueue.h
```diff
@@ -47,6 +47,9 @@ void      *taosAllocateQitem(int size);
 void       taosFreeQitem(void *item);
 int        taosWriteQitem(taos_queue, int type, void *item);
 int        taosReadQitem(taos_queue, int *type, void **pitem);
+// special type search Qitem
+int        taosSearchQitem(taos_queue, int type, void **pitem);
```
hjxilinx, line 51: search in queue?

src/client/src/tscSql.c
DuanKuanJun: Deleted 1000 row(s) from 1000 table(s) (0.331877s) .  1000 tables can be got by taos_affected_tables() api

src/util/inc/tqueue.h
DuanKuanJun: yes , need remove Qitem from queue after async delete all rows from disk.

DuanKuanJun:
this is test branch , close PR

PR #11522, opened 2022-04-15
docs: update 02-planning.mdx

boxiaobj:
modify memory requirement calculation.


gccgdb1234:
branch name for document change should be like "docs/xxxx" , the prefix "docs/" will prevent this PR from going through CI.

guanshengliang (approved):

PR #11596, opened 2022-04-18
fix(query): fix tag col check with groupby expr

stephenkgu:


stephenkgu:
same as #11345 for 2.4 branch

PR #11620, opened 2022-04-19
fix(driver): fix preparedStatement preprocessing error

huolibo:
sql will cut out if there is "* is null"
TS-1432


sangshuduo (changes requested):
if the code lines are commented off is decided to remove, just remove them.

dingbo8128 (approved):
LGTM

sangshuduo (approved):
LGTM

PR #11704, opened 2022-04-21
docs: update pages for some tools

gccgdb1234:


dingbo8128 (approved):
LGTM, but user will not konw how to install it after reading whole page.

PR #12055, opened 2022-04-29
test: demo for produce message to TQ and consume message from TQ

zyyang90:
test: demo for produce message to TQ and consume message from TQ


xiaopingcs (approved):
Looks good

PR #12088, opened 2022-05-01
fix: change cmake minimum required version to 3.0 in TSZ

ericyangpan:


sangshuduo (changes requested):
TSZ is a submodule. You should submit its PR to its repo separately and PR its tip commit back to TDengine.

sangshuduo (approved):
LGTM

PR #12133, opened 2022-05-05
fix: result is incorrect in case of interval query with elapsed()

xiao-yu-wang:


src/query/src/qAggMain.c
```diff
@@ -5186,7 +5186,7 @@ static bool elapsedSetup(SQLFunctionCtx *pCtx, SResultRowCellInfo* pResInfo) {
 }
 
 static int32_t elapsedRequired(SQLFunctionCtx *pCtx, STimeWindow* w, int32_t colId) {
```
glzhao89: maybe we can remove elapsedRequired function and use more general dataBlockRequired function instead? Since after this change there's no difference of adding elapsed specified version of this function.

glzhao89 (approved):

PR #12157, opened 2022-05-06
fix: remove python connector

sangshuduo:
update requirements.txt

[TD-14358]


sangshuduo:
already reviewed at develop branch

PR #12225, opened 2022-05-07
test: finish stb select column and tb select *|column

jiajingbin:
Probabilistic coredump and no ci temporarily


wu-champion (commented):
LGTM

PR #12267, opened 2022-05-09
enh(driver): change auth type to basic and add token property

huolibo:


sangshuduo (approved):
LGTM. Better to have at least one more review by a Java expert.

src/connector/jdbc/src/main/java/com/taosdata/jdbc/rs/RestfulStatement.java
```diff
@@ -95,6 +95,10 @@ public boolean execute(String sql) throws SQLException {
 
     private String getUrl() throws SQLException {
         String dbname = conn.getClientInfo(TSDBDriver.PROPERTY_KEY_DBNAME);
+        String protocol = "http";
```
dingbo8128, line 98: why not use https ?

src/connector/jdbc/src/main/java/com/taosdata/jdbc/rs/RestfulStatement.java
```diff
@@ -95,6 +95,10 @@ public boolean execute(String sql) throws SQLException {
...
+//        if (conn.isUseSsl()) {
+//            protocol = "https";
+//        }
```
dingbo8128, line 101: why not use https ?

changxiaoqiang (approved):
LGTM

src/connector/jdbc/src/main/java/com/taosdata/jdbc/rs/RestfulStatement.java
huolibo: Because TDengine not support ssl, this may mislead users.

dingbo8128 (approved):
LGTM

PR #12385, opened 2022-05-12
test: add ts filter testcases

jiajingbin:


wu-champion (changes requested):
need to add ：
current case:
1. ts in (time1, time2,...)
2. ts not in(...)
3. ts is null
4. ts is not null
5. group by ts
6. having ts filter
7. ...
err case:
1. ts like ""
2. ts match
3. ts nmacth
4. ...

hufeng-tany:
貌似l主键ts不能当做like的匹配列，我的版本是2.2.0.5

PR #12542, opened 2022-05-16
revert: cmake git submodule for2.4

sangshuduo:
[TD-15342]


sangshuduo:
conditionally merge but the case failed as following:

query/queryDiffColsTagsAndOr.py(818) failed: sql:select count(*), avg(c6), sum(c3) from (select * from qwobeljb where c1 >1 or c2 = 2 and c7 like "binar_" and c4 in (3, 5)) where c1 != 2 or c3 = 1 or t1=2 or t1=3 or c8 like "ncha_" and c9 in (true) interval(8d), elm:15 != expect_elm:17

PR #12678, opened 2022-05-18
docs: grammer and phrasing: taosBenchmark

sean-tdengine:
Fixed in page links as well. There is a lot here, a good job by whoever wrote this, definitely much better than the prior version. I will probably make further edits to the sample csv section tomorrow, I had to do some trial and error while working through it myself.


docs-en/14-reference/05-taosbenchmark.md
```diff
@@ -7,7 +7,7 @@ description: "taosBenchmark (once called taosdemo ) is a tool for testing the pe
...
+taosBenchmark (formerly taosdemo ) is a tool for testing the performance of TDengine products. taosBenchmark can test the performance of TDengine's insert, query, and subscription functions and simulate large amounts of data generated by many devices. taosBenchmark can be configured to generate user defined databases, supertables, subtables, and the time series data to populate these for performance benchmarking. The installer provides taosdemo as a soft link to taosBenchmark for compatibility with past users.
```
sangshuduo: Suggestion: better to mention "the time interval for inserting", "the number of working threads", and "be able to insert disordered data" too with some reformat as they are important features.

docs-en/14-reference/05-taosbenchmark.md
```diff
@@ -35,17 +35,17 @@ Execute the following commands to quickly experience taosBenchmark's default con
...
 
-The `-f <json file>` argument cannot be used when running taosBenchmark with command-line parameters and controlling its behavior. Users must specify all configuration parameters from the command-line. The following is an example of testing taosBenchmark writing performance using the command-line approach.
+The `-f <json file>` argument cannot be used when running taosBenchmark with command-line parameters. Users must specify all configuration parameters from the command-line. The following is an example of testing taosBenchmark writing performance using the command-line approach.
 
 ```bash
 taosBenchmark -I stmt -n 200 -t 100
 ```
 
-The above command, `taosBenchmark` will create a database named `test`, create a super table `meters` in it, create 100 sub-tables in the super table and insert 200 records for each sub-table using parameter binding.
+In the above command, `taosBenchmark` will create the default database named `test`, create the default super table named `meters`, create 100 subtables in the super table and insert 200 records for each subtable.
```
sangshuduo: '-I stmt' means explicitly specifying the "parameter binding" interface which is no taos_query() series interface. Please include the difference.

sangshuduo (commented):
Thanks so much and pretty nice work. Just commented on two changes.

docs-en/14-reference/05-taosbenchmark.md
sean-tdengine: Interesting, I have never heard of parameter binding interface writing. Parameter binding is typical for front end interfaces, as in modern web design, but I have never heard of writing to a database in this way. Is there another software package that uses this "parameter binding" interface that you can point me to?

sangshuduo: Please see the link https://docs.tdengine.com/develop/insert-data/sql-writing#insert-with-parameter-binding. Actually, MySQL supports this style interface first. TDengine implements a set of similar interfaces.

sangshuduo (approved):
LGTM

PR #12848, opened 2022-05-23
fix: failed to alloc memory when big number tables

sangshuduo:
for develop

[TD-15803]


sangshuduo:
code change already reviewed in the taos-tools repo. this PR is just to update commit.

PR #12849, opened 2022-05-23
fix: failed to alloc memory when big number tables

sangshuduo:
for 2.4

[TD-15803]


sangshuduo:
code change already reviewed in the taostools repo. this PR is just to update commmit.

PR #12858, opened 2022-05-23
fix: ignore Ref is not there to save true error

stephenkgu:


stephenkgu:
same as #12859

zhouyf605:
Is this a bugfix？

PR #12870, opened 2022-05-23
fix: update taos-tools to support bignum subtable for2.4

sangshuduo:
[TD-15803]


sangshuduo:
code change already reviewed in the taostools repo. this PR is just to update commmit.

PR #12900, opened 2022-05-24
docs: grammar: Connect to TDengine

sean-tdengine:
Made some grammar and reworded the differences between connectors.


sangshuduo (approved):
LGTM

PR #12958, opened 2022-05-25
Update .gitmodules

omezro:
some error happens when use "git submodule update --init --recursive" to clone submodules , and then I found some url of submodule is not a git clone url end with .git


sangshuduo (approved):
LGTM

PR #12994, opened 2022-05-25
fix the docs links

fansehep:
Some links to documentation were broken, I fixed it.
And the README.md has a error.


sangshuduo:
@fansehep please ignore the link 'issue' in the docs-cn/docs-en directory. we use a backend program to generate the official website according to the markdown there. But the second commit is definitely a mistake. Please close this PR and submit a new one to fix the README second commit try to fix. Thanks for your contribution.

PR #13022, opened 2022-05-26
refactor: use max int64 to replace -1 for unlimit

sangshuduo:
for 3.0

[TD-15999]


sangshuduo:
code change already reviewed in the taostools repo. this PR is just to update commit.

PR #13085, opened 2022-05-27
chore: make odbc cmake file compatible with lower version cmake

sangshuduo:
[TD-16059]


sangshuduo:
already reviewed at https://github.com/taosdata/TDengine/pull/13086

PR #13086, opened 2022-05-27
chore: make odbc cmake file compatible with lower version cmake

sangshuduo:
[TD-16059]


src/connector/odbc/examples/c/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ ADD_EXECUTABLE(tcodbc main.c ../../src/todbc_log.c)
 
 IF (TD_LINUX OR TD_DARWIN)
   target_include_directories(tcodbc PRIVATE ${ODBC_INCLUDE_DIRECTORY})
-  target_link_directories(tcodbc PUBLIC ${ODBC_LIBRARY_DIRECTORY})
+  link_directories(AFTER ${ODBCINST_LIBRARY_DIRECTORY})
```
freemine: better check cmake version first and choose the preferred one
```
  IF (CMAKE_VERSION VERSION_LESS 3.13)
      link_directories(AFTER ${ODBC_LIBRARY_DIRECTORY})
  ELSE ()
      target_link_directories(todbc PUBLIC ${ODBC_LIBRARY_DIRECTORY})
  ENDIF ()
```

freemine (changes requested):

src/connector/odbc/examples/c/CMakeLists.txt
sangshuduo: thanks. applied.

PR #13231, opened 2022-05-30
chore: add .mailmap file

ericyangpan:
Ref: https://github.com/torvalds/linux/blob/master/.mailmap

This list is used by git-shortlog to fix a few botched name translations
in the git archive, either because the author's full name was messed up
and/or not always written the same way, making contributions from the
same person appearing not to be so or badly displayed. Also allows for
old email addresses to map to new email addresses.


sangshuduo (approved):
LGTM

PR #13297, opened 2022-05-31
docs: test anchors

dingbo8128:


gccgdb1234:
There are 4 document versions, this change will make the style totally different and we have to change all of them. We don't have resource and time to do this. However, it's definitely fine if any one outside R&D volunteer to do this.

PR #13318, opened 2022-05-31
2.4.0.14 error occurs when using pow()  calculate decimal 

aaaaaadi:
image
image


gccgdb1234:
This is a wrongly generated PR.

PR #13942, opened 2022-06-17
fix: update taos-tools

zhaoyanggh:


sangshuduo (commented):
recent taos-tools already updated

PR #13992, opened 2022-06-20
fix: shell exit while ws rev is 0

zhaoyanggh:


sangshuduo (changes requested):
should exit only users want to

src/kit/shell/src/shellEngine.c
```diff
@@ -1241,7 +1241,7 @@ int wsclient_send(char *strdata, WebSocketFrameType frame) {
     payload_offset += 8;
   } else {
     fprintf(stderr, "websocket send too large data\n");
-    return -1;
+    return 1;
```
sangshuduo: What's the difference between 1 and -1? where it be checked?

zhaoyanggh: There is no difference, just make it the same as the error return in line 1277.

sangshuduo: Return a minus number is more industry practice.

src/kit/shell/src/shellEngine.c
```diff
@@ -1399,8 +1400,17 @@ void wsclient_parse_frame(SWSParser * parser, uint8_t * recv_buffer) {
       recv_buffer[i] = c[i] ^ ((unsigned char *) (&mask))[i % 4];
     }
   }
+  if (msg_opcode == 0x0 || msg_opcode == 0x1 || msg_opcode == 0x2) {
+    if (!msg_fin) {
+      printf("incomplete frame\n");
+    }
+  }
+  if (msg_opcode == 0xA) {
+    printf("get pong\n");
```
sangshuduo: Is it necessary to output to the end-user on screen? How about recording it in the log?

src/kit/shell/src/shellEngine.c
```diff
@@ -1399,8 +1400,17 @@ void wsclient_parse_frame(SWSParser * parser, uint8_t * recv_buffer) {
...
+  }
   if (msg_opcode == 0x9) {
     parser->frame = PING_FRAME;
+    printf("get ping\n");
```
sangshuduo: same concern

sangshuduo (approved):
LGTM

PR #14025, opened 2022-06-20
fix: remove useless script

gccgdb1234:


sangshuduo:
the script `packaging/tools/makepkg.sh` and few documents need modify accordingly.

PR #14066, opened 2022-06-21
fix(tsdb): 'extraRow' was set but never used (2.6)

sunshine-smdbs:
fix as issue described.


sunshine-smdbs:
fix while update = 2, the way to obtain row in IMemTable to merge.

PR #14082, opened 2022-06-21
fix: update taos-tools

sangshuduo:
add data file version check

[TD-13052]


sangshuduo:
Already reviewed in the taos-tools repo. just update the commit to TDengine.

PR #14269, opened 2022-06-26
fix: some warnings by Coverity Scan in monMain.c

fansehep:


xiangminghu (commented):
1. Besides -1, shouldn't charLen also be checked against 0? 

2. In the lines comparing against string literals such as "master", wouldn't it make more sense to just use strcmp?

guanshengliang (approved):

guanshengliang:
There is a conflict, close it

PR #14282, opened 2022-06-27
test: fix constant: add taos_keywords; add time_range_wise case

wu-champion:
time_range_wise does not completed


jiacy-jcy (approved):
LGTM

jiajingbin (commented):
LGTM

PR #14359, opened 2022-06-29
fix: install script

sangshuduo:
remove sudo from directory maniplation


sangshuduo:
already verified locally

PR #14360, opened 2022-06-29
fix: remove csudo from mkdir debug

sangshuduo:


sangshuduo:
already verified locally.

PR #14546, opened 2022-07-05
Test: fix the histogram case

wu-champion:


jiacy-jcy (commented):
maybe should add a scene for distribute query

PR #14872, opened 2022-07-13
fix: resolve memory leak on select interp... where range...

bgzhao66:


examples/c/makefile
```diff
@@ -7,7 +7,6 @@ LFLAGS = '-Wl,-rpath,/usr/local/taos/driver/' -ltaos -lpthread -lm -lrt
 CFLAGS = -O3 -g -Wall -Wno-deprecated -fPIC -Wno-unused-result -Wconversion \
 	-Wno-char-subscripts -D_REENTRANT -Wno-format -D_REENTRANT -DLINUX \
 	-Wno-unused-function -D_M_X64 -I/usr/local/taos/include -std=gnu99 \
-	-I../../../deps/cJson/inc \
```
bgzhao66, line 10: Remove unnecessary -I header dir for examples/c to avoid users' confusion

src/query/inc/qSqlparser.h
```diff
@@ -97,9 +97,11 @@ typedef struct SIntervalVal {
   SStrToken          offset;
 } SIntervalVal;
 
+typedef struct tSqlExpr tSqlExprTimestamp;
+
 typedef struct SRangeVal {
```
bgzhao66, line 102: SRangeVal is actually a range of timestamps  in tSqlExpr. Use typedef alias to indicate its intention to improve code readability.

src/query/src/qSqlParser.c
```diff
@@ -1183,6 +1183,10 @@ void destroySqlNode(SSqlNode *pSqlNode) {
   pSqlNode->fillType = NULL;
 
   tSqlExprDestroy(pSqlNode->pHaving);
+
```
bgzhao66, line 1186: The %destructor timestamp directive in the source code sql.y of Lemon parser generator doesn't suffice resource deallocation of timestamp tSqlExpr(s), which are stored in the external context pInfo. So, it is required to destroy the timestamp tSqlExpr(s) of time range here

winshining (approved):

PR #14907, opened 2022-07-14
fix: change the default maxDbs of root acct to INT16_MAX

bgzhao66:


src/mnode/src/mnodeAcct.c
```diff
@@ -215,7 +215,7 @@ static int32_t mnodeCreateRootAcct() {
   taosEncryptPass((uint8_t *)TSDB_DEFAULT_PASS, strlen(TSDB_DEFAULT_PASS), pAcct->pass);
   pAcct->cfg = (SAcctCfg){
     .maxUsers           = 128,
-    .maxDbs             = 128,
```
bgzhao66, line 218: Since the type of maxdbs field in the `log.acct_root` is SMALLINT, which is `i16`, it is to change the default value of .maxDbs as INT16_MAX here.

tests/script/unique/account/paras.sim
```diff
@@ -17,7 +17,7 @@ endi
 if $data02 != 3/128 then 
   return -1
 endi
-if $data03 != 0/128 then
```
bgzhao66, line 20: INT16_MAX is equal to 32767.

PR #14967, opened 2022-07-15
test:test case for update_iotdata

jiacy-jcy:


tests/system-test/1-insert/update_data.py
```diff
@@ -0,0 +1,223 @@
+###################################################################
```
jiajingbin, line 1: 1.stb/ctb/ntb
2.normal data->Null and Null -> normal data
3.binary/nchar increase length and update
4.batch-update
5.multi-table update
6.specified-column update

jiacy-jcy: 1-3 is OK，4-6 will  be fulfilled in another case

jiajingbin (approved):
LGTM

PR #14987, opened 2022-07-16
test: fix case for support rest API

wu-champion:


wu-champion:
lost create taosAdapter env

PR #15075, opened 2022-07-18
fix(query): fix function first/last do not support constant input

glzhao89:
TD-17400


glzhao89 (commented):
Test

source/libs/function/src/builtins.c
```diff
@@ -1423,14 +1423,6 @@ static int32_t translateIrate(SFunctionNode* pFunc, char* pErrBuf, int32_t len)
 }
 
 static int32_t translateFirstLast(SFunctionNode* pFunc, char* pErrBuf, int32_t len) {
-  int32_t numOfParams = LIST_LENGTH(pFunc->pParameterList);
```
glzhao89, line 1426: remove the check to make sure constant input can pass,

PR #15082, opened 2022-07-18
test: enh tools of crash_gen about start time when it insert datas

wenzhouwww:
Enhanced the function of crash_gen to automatically generate start time of writing records, and it will write data within 5 years


stevel032:
Great code comments from @wenzhouwww , kudos!

PR #15113, opened 2022-07-19
fix(query): fix mode function processing all null column 

glzhao89:
TD-17024


glzhao89 (commented):
fix(query): fix mode function processing all null column

PR #15127, opened 2022-07-19
fix(query): change cast function input limit from 1000->4096 bytes

glzhao89:
TD-17497


glzhao89 (commented):
change cast function input limit from 1000->4096 bytes

PR #15144, opened 2022-07-19
fix: resolve a potential nullptr dereferencing on extraRow with update=2

bgzhao66:


src/tsdb/src/tsdbRead.c
```diff
@@ -986,7 +986,9 @@ static SMemRow getSMemRowInTableMem(STableCheckInfo* pCheckInfo, int32_t order,
       return rmem;
     } else {
       pCheckInfo->chosen = CHECKINFO_CHOSEN_BOTH;
-      *extraRow = rimem;
```
bgzhao66, line 989: It is potential when there is the same key in mutable and immutable memory tables when update is set to be 2 for partial updating but extraRow is NULL.

PR #15220, opened 2022-07-20
Update 02-intro.md

chfree:
文档错字修改


arielyang (commented):
Thank you for your PR.
“即席查询” here means "ad-hoc queries" in English. That is correct.

PR #15234, opened 2022-07-21
fix: merge dup rows in client

kailixu:


kailixu:
Unexpected diff in this PR, reopen.

PR #15298, opened 2022-07-22
docs: update links in README.md to use docs.tdengine.com

danielclow:
This updates some links to use docs.tdengine.com instead of the older URL taosdata.com/en/documentation .

Note that there may not be an exact match for one or two links as the documentation structure seems to have changed since 2.0.


sangshuduo (approved):
LGTM

PR #15377, opened 2022-07-25
Added 2 acceptable error codes to crash_gen tool

stevel032:
Added these two errors as acceptable for the crash_gen tool:

```
+                0x0369, # Tag already exists
+                0x0388, # Database not exist
```


stevel032:
Thank you Wade for the help and quick response!

PR #15447, opened 2022-07-26
fix: remove excessive escaping of quotation marks with backslashes for schemaless insert in influxdb line protocol

bgzhao66:


src/client/src/tscParseLineProtocol.c
```diff
@@ -2311,10 +2301,11 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
             ret = TSDB_CODE_TSC_LINE_SYNTAX_ERROR;
             goto error;
           }
-
+          if (*cur == '"') {
```
bgzhao66, line 2304: For SML binary str encoding, the surrounding double-quotation marks used as identifying characteristics of its type will be cut off directly when being converted to type `TSDB_DATA_TYPE_BINARY` later in fn `convertSmlValueType`. So, the quotation marks within the string need not be  escaped with a backslash '\\' this time. Thus, the preceding backslash of an internal quotation mark is skipped and then stripped here.

src/client/src/tscParseLineProtocol.c
```diff
@@ -2077,6 +2078,11 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
     val_rqoute
   } val_state;
 
+  value = malloc(bufSize);
+  if (value == NULL) {
```
bgzhao66, line 2082: Since the init value of bufSize is TSDB_FUNC_BUF_SIZE(i.e. 512) which is not zero, ENOMEM has happened if the NULL pointer is returned by `malloc`.

src/client/src/tscParseLineProtocol.c
```diff
@@ -2077,6 +2078,11 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
     val_rqoute
   } val_state;
 
+  value = malloc(bufSize);
```
bgzhao66, line 2081: Use a preallocated buffer in heap instead of a post-allocated one, in order to be able  to construct the result incrementally during processing its syntax.

bgzhao66: Although it is less efficient to allocate a buffer in heap memory to process short strings than one in stack memory each time, the old behavior of code is preserved for simplicity since the taosAdapter as a stateless independent app which can be scaled horizontally easily on demand.

src/client/src/tscParseLineProtocol.c
```diff
@@ -2362,6 +2350,16 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
       }
     }
 
+    if (start < cur) {
+      if (bufSize <= len + (cur - start)) {
+        bufSize *= 2;
+        value = realloc(value, bufSize);
+      }
+      memcpy(value + len, start, cur - start);  // [start, cur)
```
bgzhao66, line 2363: Construct the result incrementally during processing its syntax.  There is no significant performance change observed here vs the old code, per a basic benchmark of performance of taosAdapter, which is the portal of ingesting data in InfluxDB line protocol.

src/client/src/tscParseLineProtocol.c
```diff
@@ -2362,6 +2350,16 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
       }
     }
 
+    if (start < cur) {
+      if (bufSize <= len + (cur - start)) {
```
bgzhao66, line 2354: Expand the buffer by a factor of 2 with `realloc`, when the size of buffer is not sufficient.

src/client/src/tscParseLineProtocol.c
```diff
@@ -2370,11 +2368,10 @@ static int32_t parseSmlValue(TAOS_SML_KV *pKV, const char **idx,
   if (len == 0 || ret != TSDB_CODE_SUCCESS) {
     free(pKV->key);
     pKV->key = NULL;
-    return TSDB_CODE_TSC_LINE_SYNTAX_ERROR;
+    ret = TSDB_CODE_TSC_LINE_SYNTAX_ERROR;
+    goto error;
   }
 
-  value = calloc(len + 1, 1);
```
bgzhao66, line 2376: A preallocated buffer of `value` ahead of time is used instead.

winshining (approved):

glzhao89 (approved):

PR #15677, opened 2022-08-02
docs: add zorm_example_td_test

ZHOUAXING927:
add zorm_example_td_test.go
Example of zorm using tdengine.


sangshuduo:
I think we should keep the TDengine source code repository storing the database kernel only and as light as possible. The ORM framework is most likely an independent ecosystem software that should not be included in the TDengine repo.

PR #15694, opened 2022-08-02
Adding 0x2662 as an alternative acceptable error for table-not-found condition

stevel032:
Traditionally, when a table is not found in an operation, the system (taosd) reports 0x2603 as the error code. Now we are seeing (per TD-17768) that sometimes the error code reported can also be 0x2662.

This PR is submitted to add this additional error code as acceptable under testing conditions.


stevel032:
Hey @wenzhouwww can you help me figure out why the CI is failing? the link above (http://ci.bl.taosdata.com:8080/job/NewTest/job/PR-15694/1/display/redirect) is actually a bad one.

wenzhouwww:
> Hey @wenzhouwww can you help me figure out why the CI is failing? the link above (http://ci.bl.taosdata.com:8080/job/NewTest/job/PR-15694/1/display/redirect) is actually a bad one.

Perhaps  there are unstable cases in CI , I think crash_gen code changes will not affect CI  cases , This situation is more of an occasional use case failure, and the use case needs to be strengthened .

PR #15717, opened 2022-08-03
docs(driver):update nodejs reference 3.0 in CN

xleili:


docs/examples/node/restexample/connect.js
```diff
@@ -17,4 +17,4 @@ test();
 // output:
 // server_version() |
 // ===================
-// 2.4.0.12         |
+// 3.0.0            |
```
sangshuduo: should be 3.0.0.0

docs/examples/node/nativeexample/connect.js
```diff
@@ -1,13 +1,20 @@
...
+  }
+}
+test();
 
-// run with: node connect.js
 // output:
-// Successfully connected to TDengine
+// server_version() |
+// ===================
+// 3.0.0            |
```
sangshuduo: should be 3.0.0.0

sangshuduo (approved):
LGTM, just two trivial typo

PR #15891, opened 2022-08-09
fix: count the number of aligned time intervals properly

bgzhao66:


src/os/src/detail/osTime.c
```diff
@@ -557,15 +557,23 @@ int64_t taosTimeSub(int64_t timeStamp, int64_t duration, char unit, int32_t prec
   return (int64_t)(mktime(&tm) * TSDB_TICK_PER_SECOND(precision));
 }
 
-
 int32_t taosTimeCountInterval(int64_t skey, int64_t ekey, int64_t interval, char unit, int32_t precision) {
```
bgzhao66, line 560: The taosTimeCountInterval function is to count the number of aligned local time intervals to cover the closed range of time `[skey, ekey]`.

bgzhao66, line 560: The taosTimeCountInterval function is to count the number of aligned local time intervals to cover the closed range of time `[skey, ekey]`.

src/os/src/detail/osTime.c
```diff
@@ -557,15 +557,23 @@ int64_t taosTimeSub(int64_t timeStamp, int64_t duration, char unit, int32_t prec
...
     skey = tmp;
   }
+
+#ifdef _MSC_VER
+#if _MSC_VER >= 1900
+  int64_t timezone = _timezone;
+#endif
+#endif
+
+  int64_t tz_offset = -1 * timezone * TSDB_TICK_PER_SECOND(precision);
```
bgzhao66, line 573: The alignment of time intervals (e.g. days in seconds) is affected by the local timezone.

src/os/src/detail/osTime.c
```diff
@@ -557,15 +557,23 @@ int64_t taosTimeSub(int64_t timeStamp, int64_t duration, char unit, int32_t prec
...
+
   if (unit != 'n' && unit != 'y') {
-    return (int32_t)((ekey - skey) / interval);
```
bgzhao66, line 568: The old computation is just to compute the length of the time range of [s, e] divided by the size of interval, rather than the number of covering aligned time intervals as required. So it caused violation of an assertion.

src/os/src/detail/osTime.c
bgzhao66: For example, if the closed time range of [skey, ekey] represented as ASCII time string is ['2020-01-01 23:0:0', '2020-01-02 1:0:0'] and the interval is 1day. Then the number of days to cover the closed time range [skey, ekey] is 2, instead of 1 as computed by the original code.

src/query/src/qFill.c
```diff
@@ -463,7 +463,6 @@ void taosFillSetInputDataBlock(SFillInfo* pFillInfo, const SSDataBlock* pInput)
         pFillInfo->interval.sliding,
         pFillInfo->interval.slidingUnit,
         pFillInfo->precision);
-      numOfRes += 1;
```
bgzhao66, line 466: Remove unnecessary code.

bgzhao66 (commented):
Post comments to explain code change.

PR #16074, opened 2022-08-12
modify CMakeLists.txt

freemine:
modify CMakeLists.txt to prevent `upx` from being untared at $HOME directory


sangshuduo (changes requested):
on purpose to download upx to ~/ to share among multiple projects

sangshuduo (approved):
LGTM

PR #16113, opened 2022-08-15
test: irate update

wenzhouwww:


wu-champion (commented):
It is not a good practice to modify the deviation standard, but the CI is passed, which proves that it is valid for this case. I hope that it can be changed to a more general check method in the future.

wu-champion (approved):

PR #16140, opened 2022-08-16
fix(query): check the upper boundary of startPos in hashIntervalAgg

bgzhao66:


src/query/src/qExecutor.c
```diff
@@ -1541,7 +1541,7 @@ static void hashIntervalAgg(SOperatorInfo* pOperatorInfo, SResultRowInfo* pResul
   while (1) {
     int32_t prevEndPos = (forwardStep - 1) * step + startPos;
     startPos = getNextQualifiedWindow(pQueryAttr, &nextWin, &pSDataBlock->info, tsCols, binarySearchForKey, prevEndPos);
-    if (startPos < 0) {
+    if (startPos < 0 || startPos >= pSDataBlock->info.rows) {
```
bgzhao66, line 1544: Break the loop if the `startPos` is not less than the number of retrieved rows ahead of time, which means that it had finished processing.

bgzhao66: This case might happen when `startPos = prevPosition + factor` for forwarding one step each time. And it means that it has finished processing when going out of the range by moving forwards or backwards .

DuanKuanJun (approved):

winshining (approved):

PR #16182, opened 2022-08-17
fix: remove rust-bindings, use official connector instead

zitsen:
Closes [TD-18455](https://jira.taosdata.com:18080/browse/TD-18455)


plum-lihui (commented):
lgmt

PR #16351, opened 2022-08-23
docs:update test report of English version

wenzhouwww:
update test report of English version


danielclow (approved):
I agree, because we do not have an updated performance test at this time, we should just remove the link for now. We can add it back in the future if necessary.

PR #16400, opened 2022-08-25
docs:add aptremove of tdengine for en document

jiacy-jcy:


sangshuduo:
libevent-core-2.1-7 libevent-pthreads-2.1-7 libopts25 sntp
==
TDengine does not depends on any package. The above msg should be shown due to other package removed earlier

gccgdb1234:
I removed improper lines, which should be caused by the specific testing environment,  and committed. @sangshuduo

PR #16448, opened 2022-08-26
enh: use simple hash to save agg result

kailixu:


kailixu:
TBD - more check and test

PR #16459, opened 2022-08-27
enh: use simple hash to save agg result

kailixu:


kailixu:
Close PR as to make CI re-schedule

PR #16484, opened 2022-08-29
docs: change TAOS SQL to TDengine SQL

ericyangpan:


docs/en/14-reference/13-schemaless/13-schemaless.md
```diff
@@ -61,14 +61,14 @@ Note that if the wrong case is used when describing the data type suffix, or if
...
 Schemaless writes process row data according to the following principles.
 
-1. You can use the following rules to generate the subtable names: first, combine the measurement name and the key and value of the label  into the next string:
+1. You can use the following rules to generate the subtable names: first, combine the measurement name and the key and value of the label into the next string:
 
 ```json
 "measurement,tag_key1=tag_value1,tag_key2=tag_value2"
 ```
 
-Note that tag_key1, tag_key2 are not the original order of the tags entered by the user but the result of using the tag names in ascending order of the strings. Therefore, tag_key1 is not the first tag entered in the line protocol.
```
ericyangpan: this need to be fixed :-(

PR #16648, opened 2022-09-04
fix(rpc): avoid except

yihaoDeng:


guanshengliang:
Already committed to 3.0 branch, no need to merge to main branch if we don't build a hotfix version

PR #16718, opened 2022-09-07
fix: resolve an assertion of index out of range in subAndCheckDone

bgzhao66:


src/client/src/tscSubquery.c
```diff
@@ -101,13 +101,12 @@ static bool allSubqueryDone(SSqlObj *pParentSql) {
 
 bool subAndCheckDone(SSqlObj *pSql, SSqlObj *pParentSql, int idx) {
   SSubqueryState *subState = &pParentSql->subState;
-  assert(idx < subState->numOfSub);
```
bgzhao66, line 104: The index may be equal to or larger than subState->numOfSub, due to subqueries (pSubs) being destroyed from another thread concurrently.

src/client/src/tscSubquery.c
```diff
@@ -101,13 +101,12 @@ static bool allSubqueryDone(SSqlObj *pParentSql) {
...
 
   pthread_mutex_lock(&subState->mutex);
-
-  tscDebug("0x%"PRIx64" subquery:0x%"PRIx64", index:%d state set to 1", pParentSql->self, pSql->self, idx);
-  subState->states[idx] = 1;
-
+  if (idx < subState->numOfSub) {
```
bgzhao66, line 106: It doesn't change the behavior of code on normal conditions, but just ignores the case of index being out of range on concurrent destruction of subqueries, since concurrent accessing to statuses of subqueries happens commonly.

bgzhao66 (commented):
Post comments

dapan1121 (approved):

guanshengliang (approved):

PR #16768, opened 2022-09-09
fix: change the size limit of sql stmt rebuilt for show create table to TSDB_MAX_SQL_LEN

bgzhao66:


cmake/define.inc
```diff
@@ -224,9 +224,9 @@ IF (TD_DARWIN_64)
   ADD_DEFINITIONS(-DUSE_LIBICONV)
   MESSAGE(STATUS "darwin64 is defined")
   IF ("${CPUTYPE}" STREQUAL "apple_m1")
-    SET(COMMON_FLAGS "-Wall -Werror -Wno-missing-braces -fPIC -D_FILE_OFFSET_BITS=64 -D_LARGE_FILE")
+    SET(COMMON_FLAGS "-Wall -Werror -Wno-missing-braces -Wno-tautological-constant-out-of-range-compare -Wno-unused-but-set-variable -fPIC -D_FILE_OFFSET_BITS=64 -D_LARGE_FILE")
```
bgzhao66, line 227: To be consistent on compiler warnings among Darwin, Linux, ARM64 and Win OSes at the this point. 
In OSes other than Darwin, the compilers (i.e. GCC in Linux and so) ignores the tautological assert statements which are always true without complaining.

bgzhao66: In Darwin, Clang's compiler warning on unused-but-set-variable is not accurate. Disable it too.

Jenkinsfile2
```diff
@@ -240,12 +240,12 @@ def pre_test_win(){
     }
     bat '''
         cd %WIN_INTERNAL_ROOT%
-        git pull
+        git pull  origin  ''' + env.CHANGE_TARGET + '''
```
bgzhao66, line 243: It is to avoid the git pull issue during CI, submitted by Haoran.

src/client/src/tscLocal.c
```diff
@@ -295,7 +295,7 @@ void tscSCreateCallBack(void *param, TAOS_RES *tres, int code) {
     taos_fetch_rows_a(tres, tscSCreateCallBack, param);    
     builder->callStage = SCREATE_CALLBACK_RETRIEVE;
   } else {
-    char *result = calloc(1, TSDB_MAX_BINARY_LEN);
+    char *result = calloc(1, TSDB_MAX_SQL_LEN);
```
bgzhao66, line 298: Use TSDB_MAX_SQL_LEN to control the length of rebuilt sql stmt, instead of TSDB_MAX_BINARY_LEN.

dapan1121 (approved):

Issue #14650, opened 2022-07-07
parsing conflicts exist in sql.y in 3.0 branch


freemine:
**Bug Description**
There 2 parsing conflicts if we compile sql.y with `lemon`, which's believed to be the corresponding sql-parser-generator used internally by TDengine, though it's not included in the source base.

**To Reproduce**
1. lemon -d/tmp/ source/libs/parsers/inc/sql.y

**Expected Behavior**
A working-and-correct grammar (no matter it's SQL or any other `language`) shall not generate parsing conflict, otherwise,
it'll result in non-deterministic-or-correct result during run-time because of ambiguities therein.

**Screenshots**

**Environment (please complete the following information):**
 - OS: [Linux xxh-VirtualBox 5.13.0-51-generic #58~20.04.1-Ubuntu SMP Tue Jun 14 11:29:12 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux; Ubuntu 20.04 LTS]
 - Memory, CPU, current Disk Space
 - TDengine Version [3.0, git commit 886af82c57e481378ad8161ddd1e2fb73c44032b]
 - lemon: (lemon -x): Lemon version 1.0

**Additional Context**
If you check the detail in /tmp/sql.out, you'll probablly find something like this:
```
State 164:
    (133) specific_cols_opt ::= * 
          specific_cols_opt ::= * NK_LP col_name_list NK_RP 
          cmd ::= INSERT INTO full_table_name * specific_cols_opt query_expression

                         NK_LP shift        80     
                         NK_LP reduce       133     ** Parsing conflict **                                                             
             specific_cols_opt shift        53     
                     {default} reduce       133    specific_cols_opt ::=
```
and
```
State 564:
    (458) query_expression ::= query_expression_body order_by_clause_opt slimit_clause_opt limit_clause_opt *
          query_primary ::= NK_LP query_expression_body order_by_clause_opt slimit_clause_opt limit_clause_opt * NK_RP
          
                         NK_RP shift-reduce 463    query_primary ::= NK_LP query_expression_body order_by_clause_opt slimit_clause_opt limit_clause_opt NK_RP   
                         NK_RP reduce       458     ** Parsing conflict **                                                             
```

It's believed that the parsing conflicts root in `query_expression rule-set'.

xiao-yu-wang:
The purpose of this rule is to hopefully implement the syntax:
[...]

PR #16849, opened 2022-09-14
fix sql.y conflicts

freemine:
https://github.com/taosdata/TDengine/issues/14650

just for demonstration,
this new syntax has no action code.


freemine:
https://github.com/taosdata/TDengine/pull/16891

Issue #16963, opened 2022-09-20
example lua51 connector build with warning, run got error!


cdnplus:
**Bug Description**
pull the main branch code,  then build connector file "examples/lua/lua51/lua_connector51.c", got some warning message.
Detail:

> lua_connector51.c: In function ‘l_open_stream’:
lua_connector51.c:313:3: warning: implicit declaration of function ‘taos_open_stream’ [-Wimplicit-function-declaration]
   void* s = taos_open_stream(taos,sqlstr,stream_cb,stime,p,NULL);
   ^
lua_connector51.c:313:13: warning: initialization makes pointer from integer without a cast [enabled by default]
   void* s = taos_open_stream(taos,sqlstr,stream_cb,stime,p,NULL);
             ^
lua_connector51.c: In function ‘l_close_stream’:
lua_connector51.c:340:3: warning: implicit declaration of function ‘taos_close_stream’ [-Wimplicit-function-declaration]
   taos_close_stream(p->stream);
   ^

got file "luaconnector51.so". run command **lua test.lua** got error message.
Detail:

> lua: error loading module 'luaconnector51' from file './luaconnector51.so':
        ./luaconnector51.so: undefined symbol: taos_open_stream
stack traceback:
        [C]: ?
        [C]: in function 'require'
        test.lua:1: in main chunk
        [C]: ? 

**To Reproduce**
2. Type in '....'
3. See error

**Environment (please complete the following information):**
 - OS: CentOS Linux release 7.7.1908 (Core)
 - Memory, CPU, current Disk Space
 - TDengine Version 3.0.1.1

sangshuduo:
Thanks for your reporting. The PR https://github.com/taosdata/TDengine/pull/16979 will fix it.

cdnplus:
thanks your reply.But I pull the fix code, build it with no warning, running test.lua then got error.
Detail:

> lua: error loading module 'luaconnector51' from file './luaconnector51.so':
        ./luaconnector51.so: undefined symbol: luaL_setfuncs
stack traceback:
        [C]: ?
        [C]: in function 'require'
        test.lua:1: in main chunk
        [C]: ?

- Lua version: 5.1.4

sangshuduo:
我用 Lua 5.3.3 (Ubuntu 20.04) 执行 build.sh 然后 lua test.lua 没问题

cdnplus:
I want to use openresty with lua, openresty only support lua 5.1

robotspace:
[...]

PR #16979, opened 2022-09-21
fix: remove stream api and use tmq syntax in lua example

sangshuduo:


sangshuduo:
[#16963](https://github.com/taosdata/TDengine/issues/16963)

PR #17044, opened 2022-09-23
feat:add new interface for schemaless to support '\0' in line value

wangmm0220:


src/inc/taos.h
```diff
@@ -215,6 +215,7 @@ DLL_EXPORT void      taos_unsubscribe(TAOS_SUB *tsub, int keepProgress);
 DLL_EXPORT int taos_load_table_info(TAOS *taos, const char* tableNameList);
 
 DLL_EXPORT TAOS_RES *taos_schemaless_insert(TAOS* taos, char* lines[], int numLines, int protocol, int precision);
+DLL_EXPORT TAOS_RES *taos_schemaless_insert_new(TAOS* taos, char* lines, int len, int32_t *totalRows, int protocol, int precision);
```
sangshuduo, line 218: how about use 'taos_schemaless_insert_with_len()'?

sangshuduo (commented):
commented

PR #17090, opened 2022-09-27
chore: install new web service instead of nginxd

zitsen:
Partial of [TD-19267](https://jira.taosdata.com:18080/browse/TD-19268)


sangshuduo (approved):
LGTM

PR #17129, opened 2022-09-28
docs: add _irowts pseudocolumn description in ZN/EN docs

glzhao89:


docs/en/12-taos-sql/06-select.md
```diff
@@ -181,6 +181,14 @@ In TDengine, the first column of all tables must be a timestamp. This column is
 select _rowts, max(current) from meters;
 ```
 
+**\_IROWTS**
+
+The \_IROWTS pseudocolumn can only be used with INTERP function. This pseudocolumn can be used to retrieve the corresponding timestamp column associated with the interpolation results.
```
gccgdb1234, line 186: pseudocolumn 少个空格吧

PR #17259, opened 2022-10-10
fix memory leakage, when parser fails to parse insert sql in some cases

freemine:
pseudo code:
```
stmt_prepare(insert into ? using st tags(?) values (?));
stmt_get_tag_fields(....);
// this is expected to fail, because stmt_set_tbname has not been called in advance
// but internally, some vgroupshashobj/tableblockhashobj is left un-free
....
stmt_close();

### CODE REVIEW

**Comment by freemine** [comment]
this is not that simple to fix. see line 1704
```
      code = parseInsertBody(&context);
```
classic ownership problem, error prune.

PR #17268, opened 2022-10-10
fix: set timestamp precision of query in tscCreateQueryFromQueryInfo

bgzhao66:


src/client/src/tscUtil.c
```diff
@@ -5238,6 +5238,8 @@ int32_t tscCreateQueryFromQueryInfo(SQueryInfo* pQueryInfo, SQueryAttr* pQueryAt
 
   STableMetaInfo* pTableMetaInfo = pQueryInfo->pTableMetaInfo[0];
 
+  pQueryAttr->precision = pTableMetaInfo->pTableMeta->tableInfo.precision;
```
bgzhao66, line 5241: The  pQueryAttr->precision is used to set the precision of timestamp of SFillInfo, which is used to compute alignment of sliding windows (e.g. of unit day or week), esp. for non-UTC timezones.

guanshengliang (approved):

PR #17322, opened 2022-10-12
fix: enhance subroutines to resolve WAL inconsistency after crashing or disk full

bgzhao66:


include/os/osEnv.h
```diff
@@ -49,14 +49,20 @@ extern SDiskSpace tsTempSpace;
 void osDefaultInit();
 void osUpdate();
 void osCleanup();
+
 bool osLogSpaceAvailable();
 bool osDataSpaceAvailable();
 bool osTempSpaceAvailable();
+
```
bgzhao66, line 56: We classify disk space conditions into two different cases : 1. disk space is in shortage, but not full yet (i.e. sufficiency). 2. disk full, no space left on disk at all (i.e. availability).

source/client/src/clientMain.c
```diff
@@ -146,21 +146,20 @@ void taos_close(TAOS *taos) {
 
 int taos_errno(TAOS_RES *res) {
   if (res == NULL || TD_RES_TMQ_META(res)) {
-    if (terrno == TSDB_CODE_RPC_REDIRECT) terrno = TSDB_CODE_RPC_NETWORK_UNAVAIL;
+    if (terrno == TSDB_CODE_RPC_REDIRECT) terrno = TSDB_CODE_QRY_NOT_READY;
```
bgzhao66, line 149: It is misleading to report RPC_REDIRECT as network unavailability. So change it.

source/dnode/mgmt/node_mgmt/src/dmEnv.c
```diff
@@ -51,26 +51,14 @@ static int32_t dmInitMonitor() {
 
 static bool dmCheckDiskSpace() {
   osUpdate();
-  if (!osDataSpaceAvailable()) {
-    dError("free disk size: %f GB, too little, require %f GB at least at least , quit",
-           (double)tsDataSpace.size.avail / 1024.0 / 1024.0 / 1024.0,
-           (double)tsDataSpace.reserved / 1024.0 / 1024.0 / 1024.0);
-    terrno = TSDB_CODE_NO_AVAIL_DISK;
-    return false;
+  if (!osDataSpaceSufficient()) {
```
bgzhao66, line 54: Still allow taosd to start on condition of disk shortage. At the same time, print some error messages. If possible, it is better to trigger real-time alarm at appropriate intervals relevant to disk usage of space. To allow OAM team to take action to prevent from hitting the condition of disk full.

source/dnode/vnode/src/tsdb/tsdbMemTable.c
```diff
@@ -160,6 +160,7 @@ int32_t tsdbDeleteTableData(STsdb *pTsdb, int64_t version, tb_uid_t suid, tb_uid
     goto _err;
   }
 
+  ASSERT(pPool != NULL);
```
bgzhao66, line 163: Since that It has not been implemented to recover from a disk error during vnodeCommit while taosd keeps running, (despite immediate crashing  of taosd may have been eliminated for disk writing or flushing), and  if unfortunately, at the same time, there are some writing requests sneaking, (although writing requests are rejected at the entrance if disk space not available  anymore),  it is better to abort with an assert of certain internal invariant of logic, instead of segfault.

source/libs/sync/src/syncMain.c
```diff
@@ -2733,7 +2736,7 @@ int32_t syncNodeOnClientRequestCb(SSyncNode* ths, SyncClientRequest* pMsg, SyncI
     code = ths->pLogStore->syncLogAppendEntry(ths->pLogStore, pEntry);
     if (code != 0) {
       // del resp mgr, call FpCommitCb
-      ASSERT(0);
+      sError("vgId:%d, failed to append log entry since %s",  ths->vgId, tstrerror(terrno));
```
bgzhao66, line 2739: Asserts replaced with error codes and relevant log messages in a lot of similar places, esp. for error handling conditions.  However, it may be all right to place an assertion on some internal invariant of logic, to detect the errors of logic assumptions alarmingly, which the correctness of code is relied upon.

source/libs/sync/src/syncRaftLog.c
```diff
@@ -414,13 +412,11 @@ int32_t logStoreAppendEntry(SSyncLogStore* pLogStore, SSyncRaftEntry* pEntry) {
   return 0;
 }
 
-SSyncRaftEntry* logStoreGetEntry(SSyncLogStore* pLogStore, SyncIndex index) {
+SSyncRaftEntry* logStoreGetEntryWithoutLock(SSyncLogStore* pLogStore, SyncIndex index) {
   SSyncLogStoreData* pData = pLogStore->data;
   SWal*              pWal = pData->pWal;
 
   if (index >= SYNC_INDEX_BEGIN && index <= logStoreLastIndex(pLogStore)) {
-    taosThreadMutexLock(&(pData->mutex));
```
bgzhao66, line 422: It is error prone to handling resource cleanup (e.g. lock acquirement/release) neatly in nested conditions. It may be better to split it into subroutines when possible.

source/libs/tdb/src/db/tdbPager.c
```diff
@@ -590,23 +613,38 @@ int tdbPagerRestore(SPager *pPager, SBTree *pBt) {
...
     if (ret < 0) {
-      ASSERT(0);
+      tdbError("failed to write buf due to %s. file: %s, bufsize:%d", strerror(errno), pPager->dbFileName, pPager->pageSize);
+      terrno = TAOS_SYSTEM_ERROR(errno);
       return -1;
     }
   }
 
-  tdbOsFSync(pPager->fd);
+  if (tdbOsFSync(pPager->fd) < 0) {
```
bgzhao66, line 629: For the handling of system errors, there are three different scenarios. 1. just ignore, (i.e. prefixed it with (void) to indicate intentionality) 2. reporting and return error code. 3. further more on 2.( i.e. report, and return error code), recover from it additionally. It is better to indicate explicitly the intentional treatment of system error in the current context.

source/libs/wal/src/walMgmt.c
```diff
@@ -98,15 +104,14 @@ SWal *walOpen(const char *path, SWalCfg *pCfg) {
   tstrncpy(pWal->path, path, sizeof(pWal->path));
   if (taosMkDir(pWal->path) != 0) {
     wError("vgId:%d, path:%s, failed to create directory since %s", pWal->cfg.vgId, pWal->path, strerror(errno));
-    taosMemoryFree(pWal);
-    return NULL;
+    goto _err;
```
bgzhao66, line 107: There is no evil to use goto _err for handling resource cleanup. In the C programming which is lacking of RAII, it is neatest way to deal with resource acquisition/cleanup with it when used properly.

source/libs/wal/src/walSeek.c
```diff
@@ -79,13 +79,23 @@ int64_t walChangeWrite(SWal* pWal, int64_t ver) {
   TdFilePtr pIdxTFile, pLogTFile;
   char      fnameStr[WAL_FILE_LEN];
   if (pWal->pLogFile != NULL) {
+    code = taosFsyncFile(pWal->pLogFile);
```
bgzhao66, line 82: fsync log and idx file before switching the file to append.

source/libs/wal/src/walMeta.c
```diff
@@ -35,113 +35,251 @@ int64_t FORCE_INLINE walGetCommittedVer(SWal* pWal) { return pWal->vers.commitVe
...
+static FORCE_INLINE int walBuildMetaName(SWal* pWal, int metaVer, char* buf) {
+  return sprintf(buf, "%s/meta-ver%d", pWal->path, metaVer);
 }
 
-static FORCE_INLINE int64_t walScanLogGetLastVer(SWal* pWal) {
+static FORCE_INLINE int walBuildTmpMetaName(SWal* pWal, char* buf) {
+  return sprintf(buf, "%s/meta-ver.tmp", pWal->path);
+}
+
+static FORCE_INLINE int64_t walScanLogGetLastVer(SWal* pWal, int32_t fileIdx) {
```
bgzhao66, line 46: There are  three different types of WAL files, i) meta info, ii) log files; iii) idx files, which are required to be in consistent condition almost always, e.g. especially on conditions of system or app crashing, disk fault, or file corruptions. It is a complicated issue to deal with since the three different types of files cannot be written into disk atomically, let alone all the three simultaneously, while confronting with various kinds of system failures, faults, or crashing, which is inevitably resulted in inconsistent states of different types of them. So it is vital to resolve these inconsistent states of files after crashing or fault. 
The original routines of inconsistency resolution were still incomplete, some already noted in scattered comments within, some not yet, which are walScanLogGetLastVer, walCheckAndRepairMeta, and walCheckAndRepairIdx.  
So, they are refactored, with adjustment of walSaveMeta fsync policies and so on, hoping to be able to handle various unresolved problems of WAL recovery, esp. on customer sites.  
BTW, the proper functioning of Raft subsystem has direct dependency on the WAL subsystem, esp. for persistence of log history, reflected by availability of the snapshotIndex, appliedIndex, commitIndex, lastLogIndex,  and  the content within that range.

source/libs/wal/src/walMeta.c
```diff
@@ -35,113 +35,251 @@ int64_t FORCE_INLINE walGetCommittedVer(SWal* pWal) { return pWal->vers.commitVe
...
+    wWarn("vgId:%d, appliedVer:%" PRId64 " is less than snapshotVer:%" PRId64 ". align with it.", pWal->cfg.vgId,
+          pWal->vers.appliedVer, pWal->vers.snapshotVer);
+    pWal->vers.appliedVer = pWal->vers.snapshotVer;
+  }
+
+  pWal->vers.commitVer = TMIN(pWal->vers.lastVer, pWal->vers.commitVer);
+  pWal->vers.appliedVer = TMIN(pWal->vers.commitVer, pWal->vers.appliedVer);
 }
 
 int walCheckAndRepairMeta(SWal* pWal) {
```
bgzhao66, line 285: It is refactored.

source/libs/wal/src/walMeta.c
```diff
@@ -184,225 +321,237 @@ int walCheckAndRepairMeta(SWal* pWal) {
...
+  if (walValidHeadCksum(pCkHead) != 0) {
+    terrno = TSDB_CODE_WAL_CHKSUM_MISMATCH;
+    return -1;
   }
 
   return 0;
 }
 
-int walCheckAndRepairIdx(SWal* pWal) {
+int walCheckAndRepairIdxFile(SWal* pWal, int32_t fileIdx) {
```
bgzhao66, line 420: It is refactored.

source/libs/wal/src/walMeta.c
```diff
@@ -565,29 +730,75 @@ static int walFindCurMetaVer(SWal* pWal) {
 int walSaveMeta(SWal* pWal) {
```
bgzhao66, line 730: It is refactored.

source/libs/wal/src/walMeta.c
```diff
@@ -35,113 +35,251 @@ int64_t FORCE_INLINE walGetCommittedVer(SWal* pWal) { return pWal->vers.commitVe
...
+
+  taosArrayClear(metaLogList);
+
+  for (int i = 0; i < actualFileNum; i++) {
+    SWalFileInfo* pFileInfo = taosArrayGet(actualLogList, i);
+    taosArrayPush(metaLogList, pFileInfo);
+  }
+}
+
+void walAlignVersions(SWal* pWal) {
```
bgzhao66, line 259: Align versions of WAL logs. i.e. indexes of raft logs, after recovering, to maintain inner consistency between them.

bgzhao66 (commented):
Post comments.

bgzhao66:
Notes: After merging 3.0, syncIsReadyForRead is removed, so it is still not allowed to read a non-leader node.

It means that the _taosd_ may not be able to be queried or even to be connected by _taos_ shell on restarting with disk full, since the raft subsystem cannot reach its leader state on failing to append a NOOP log entry after leader election.

guanshengliang (approved):

PR #17677, opened 2022-10-26
fix(coverage): remove the function about skiplist callback

DuanKuanJun:


DuanKuanJun:
I had confirmed this function with Jichong.Liu and HongZeng.Chen ,  It was obsolete that callback when inserting skiplist.
So delete .

PR #17732, opened 2022-10-28
fix: use the relative pos during searching in walScanLogGetLastVer

bgzhao66:


source/libs/wal/src/walMeta.c
```diff
@@ -173,19 +179,19 @@ static FORCE_INLINE int64_t walScanLogGetLastVer(SWal* pWal, int32_t fileIdx) {
       }
 
       // found one
-      found = candidate;
-      haystack = candidate + 1;
+      found = pos;
```
bgzhao66, line 182: The memory address might change if realloc() obtains a different location of memory for the internal buffer. Use a relative position for found result.

bgzhao66 (commented):
post comments

cosmic-jc (approved):

guanshengliang (approved):

PR #17824, opened 2022-11-02
fix: set firstTrial as false if error encountered in loop of walScanLogGetLastVer

bgzhao66:


bgzhao66:
Fixed an issue: if a bad record appears in WAL just after offset last fsync, it may be skipped in the log, causing a bad record sneaking in.

source/libs/wal/src/walMeta.c
```diff
@@ -201,7 +201,13 @@ static FORCE_INLINE int64_t walScanLogGetLastVer(SWal* pWal, int32_t fileIdx) {
     }
 
     if (end == fileSize) firstTrial = false;
-    if (firstTrial && terrno == TSDB_CODE_SUCCESS) continue;
+    if (firstTrial) {
```
bgzhao66, line 204: If no valid WAL record is detected yet in scanning forward while encountering an error, e.g. CHSUM mismatch, then scan backwards. In this time, set firstTrial as false, to prevent it from continuing to scan forwards.

bgzhao66:
The chance of triggering this bug is low.

cosmic-jc (approved):

guanshengliang (approved):

PR #17879, opened 2022-11-03
fix: taosbenchmark query interval unit

sangshuduo:


sangshuduo:
same change as #17847

PR #17980, opened 2022-11-09
feat: taosbenchmark support query fails then continue

sangshuduo:


sangshuduo:
already passed test in taos-tools github actions

PR #17981, opened 2022-11-09
feat: taosbenchmark support query fails then continue for 2.6

sangshuduo:


sangshuduo:
already test passed in taos-tools repo

PR #18227, opened 2022-11-17
fix: real path optimization in case of invalid read

kailixu:


kailixu:
Should use 3.0 branch.

PR #18345, opened 2022-11-22
fix:  crash caused by elect null and issues found in asan

dapan1121:


source/libs/executor/src/sortoperator.c
```diff
@@ -377,10 +377,10 @@ int32_t beginSortGroup(SOperatorInfo* pOperator) {
   param->childOpInfo = pOperator->pDownstream[0];
   param->grpSortOpInfo = pInfo;
   ps->param = param;
+  ps->onlyRef = false;
   tsortAddSource(pInfo->pCurrSortHandle, ps);
 
   int32_t code = tsortOpen(pInfo->pCurrSortHandle);
-  taosMemoryFreeClear(ps);
```
shenglian-zhou, line 383: when to free ps?
I found it then

shenglian-zhou (approved):

PR #18363, opened 2022-11-22
refactor(sync): do not merge this PR!!   just to find which case can not pass.

castermode:


guanshengliang:
You can set the PR to draft to avoid merging

PR #18374, opened 2022-11-22
fix: clear the obsolete data blocks when put data block

kailixu:


kailixu:
Already fixed in #18362

PR #18635, opened 2022-12-02
doc: update grafana version in English doc and modifythe way grafana plugin installed in application doc

sunpe:
…plugin installed in application doc


sangshuduo (approved):
LGTM

PR #18701, opened 2022-12-05
fix: walSaveMeta ahead of deleting WAL files as gc in walEndSnapshot

bgzhao66:


source/libs/wal/src/walMgmt.c
```diff
@@ -121,7 +121,16 @@ SWal *walOpen(const char *path, SWalCfg *pCfg) {
...
+           pWal->path);
+    goto _err;
+  }
+
+  // init gc
+  pWal->toDeleteFiles = taosArrayInit(8, sizeof(SWalFileInfo));
+  if (pWal->toDeleteFiles == NULL) {
+    wError("vgId:%d, failed to init taosArray of toDeleteFiles due to %s. path:%s", pWal->cfg.vgId, strerror(errno),
+           pWal->path);
     goto _err;
```
cosmic-jc, line 134: totDeleteFiles array is not destroyed

source/libs/wal/src/walWrite.c
```diff
@@ -392,6 +376,26 @@ int32_t walEndSnapshot(SWal *pWal) {
...
+    }
+    walBuildIdxName(pWal, pInfo->firstVer, fnameStr);
+    wDebug("vgId:%d, wal remove file %s", pWal->cfg.vgId, fnameStr);
+    if (taosRemoveFile(fnameStr) < 0 && errno != ENOENT) {
+      ASSERT(0);
+    }
+  }
+  taosArrayClear(pWal->toDeleteFiles);
+
 END:
```
cosmic-jc, line 394: will it be better to put array clear after end label?

source/libs/wal/src/walWrite.c
```diff
@@ -350,34 +351,17 @@ int32_t walEndSnapshot(SWal *pWal) {
...
-    }
 
   UPDATE_META:
     // make new array, remove files
-    taosArrayPopFrontBatch(pWal->fileInfoSet, actualDelete);
+    taosArrayPopFrontBatch(pWal->fileInfoSet, deleteCnt);
     if (taosArrayGetSize(pWal->fileInfoSet) == 0) {
       pWal->writeCur = -1;
-      pWal->vers.firstVer = -1;
+      pWal->vers.firstVer = pWal->vers.lastVer + 1;
```
cosmic-jc, line 359: if all files are deleted, first version should be set to -1

cosmic-jc (approved):

source/libs/wal/src/walMgmt.c
bgzhao66: resolved.

source/libs/wal/src/walWrite.c
bgzhao66: Next time, the partial cleaned WAL files as gc are to be cleaned in next attempt.

source/libs/wal/src/walWrite.c
bgzhao66: the close-open interval [) is for an empty set.

guanshengliang (approved):

PR #19078, opened 2022-12-22
Fix/xsren/td 20817 mac fqdn

facetosea:


source/os/src/osSysinfo.c
```diff
@@ -1007,6 +1007,11 @@ SysNameInfo taosGetSysNameInfo() {
     tstrncpy(info.machine, uts.machine, sizeof(info.machine));
   }
 
+  char     localHostName[512];
+  taosGetlocalhostname(localHostName, 512);
+  TdCmdPtr pCmd = taosOpenCmd("scutil --get LocalHostName");
```
sangshuduo, line 1015: is there any interface to call for getting localhostname instead of calling external command?

sangshuduo (commented):
see my comment

sangshuduo (approved):
LGTM

PR #19522, opened 2023-01-11
docs:cdiwadkar16-patch-1-1 - Updating for parameter spelling errors

cdiwadkar16:
the parameter prepared_rand was incorrectly written as prepare_rand in 2 places. Also changed "detailed" to "detail" in 2 places for grammatical reasons.

### DIFF
--- a/docs/en/14-reference/05-taosbenchmark.md
+++ b/docs/en/14-reference/05-taosbenchmark.md
@@ -92,7 +92,7 @@ taosBenchmark -f <json file>
 
 </details>
 
-## Command-line argument in detailed
+## Command-line argument in detail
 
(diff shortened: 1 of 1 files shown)


sangshuduo:
@cdiwadkar16  thanks for your correction. the source code need to be modified accordingly too. Please notify me if you have similar request in future.

sangshuduo:
> @cdiwadkar16 thanks for your correction. the source code need to be modified accordingly too. Please notify me if you have similar request in future.↳

@cdiwadkar16 sorry my mistake, it's only a mistake in document, the code is correct so no need to modify. Thanks again for your correction.

PR #19543, opened 2023-01-12
create stream and use existing super table

54liuyao:


include/libs/qcom/query.h
```diff
@@ -207,6 +207,12 @@ typedef struct SQueryNodeStat {
   int32_t tableNum;  // vg table number, unit is TSDB_TABLE_NUM_UNIT
 } SQueryNodeStat;
 
+typedef struct SColLocation {
```
hjxilinx, line 210: If we can use the SColumn, instead of whole new struct, as it is not used to transfer message?

hjxilinx (approved):

PR #19592, opened 2023-01-17
fix stackflow on mac

facetosea:


source/os/src/osSysinfo.c
```diff
@@ -18,6 +18,7 @@
 #include "taoserror.h"
 
 #define PROCESS_ITEM 12
+#define uuidLen37 37
```
sangshuduo: should be ALL UPPER CASE for macro definition

facetosea: O, mistake that shouldn't be made. I have fixed.

sangshuduo (approved):
LGTM

PR #19658, opened 2023-01-22
some modification in make_install.bat

freemine:
1. copy files failed if svc-deleting is still outstanding
2. unnecessary duplication of C:\TDengine in %PATH%

### DIFF
--- a/packaging/tools/make_install.bat
+++ b/packaging/tools/make_install.bat
@@ -2,12 +2,20 @@
 
 for /F %%a in ('echo prompt $E ^| cmd') do set "ESC=%%a"
 
-goto %1
+if "%1" NEQ "" goto %1
+
 :needAdmin
 
 if exist C:\\TDengine\\data\\dnode\\dnodeCfg.json (
   echo The default data directory C:/TDengine/data contains old data of tdengine 2.x, please clear it before installing!
 )
+
+rem // stop and delete service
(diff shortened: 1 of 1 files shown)


packaging/tools/make_install.bat
```diff
@@ -105,3 +113,36 @@ rem // use setx to set a temporary throwaway value to trigger a WM_SETTINGCHANGE
...
+
+:stop_delete
+sc stop taosd
+sc delete taosd
+sc stop taosadapter
+sc delete taosadapter
+exit /B 0
+
+:check_svc
+sc query %1 >nul 2>nul && goto :check_svc %1
```
facetosea, line 145: Is there any situation here that will cause an infinite loop?

freemine:
please comment @sangshuduo @facetosea

facetosea:
@freemine I am worried that check_svc function may have an infinite loop, other changes I think no problem

freemine:
@facetosea, thx.

case1: there's no registered `taosd` service, `sc query taosd` would fail and break out the loop
case2: `taosd` is running, this could not happend, because `sc stop taosd` has just been called previously
case3: `taosd` is stopping, `sc query taosd` would success with status and thus, continue the loop
case4: once `taosd` is stopped, deletion of `taosd` will complete, and switch to `case1`

it seems the only way to make the loop infinite is `taosd` does not yield execution when being requested to stop, which, optimistically, is believed not to happen. :)

facetosea (approved):
LGTM

PR #19662, opened 2023-01-22
enh: perform compact only if there is benefit

kailixu:


kailixu:
commit when compact with specified time range finish.

PR #19734, opened 2023-02-01
Chore/xsren/td 19391/oem

facetosea:


packaging/tools/install_client.sh
```diff
@@ -23,6 +23,18 @@ osType=Linux
 pagMode=full
 verMode=edge
 
+clientName2="taos"
+serverName2="taosd"
+productName2="TDengine"
+emailName2="taosdata.com"
+
+adapterName2="${clientName2}adapter"
```
sangshuduo: no need to change any daemon name

sangshuduo (changes requested):
need to discuss more

Issue #19665, opened 2023-01-24
too many functions exported in libtaos.so/libtaos.dylib


freemine:
on MacOS:
```
nm -g debug/build/lib/libtaos.1.dylib
```
shows, that not only functions declared in `taos.h` are exported in dynamic library, but also `internal` functions, which's not elegant and might introduce security vulnerability in target machine.
here's some screenshots.
```
0000000000384ec0 T _vnodeProcessQueryMsg
0000000000387750 T _vnodeProcessSyncMsg
0000000000381360 T _vnodeProcessWriteMsg
0000000000385b40 T _vnodeProposeCommitOnNeed
0000000000385ea0 T _vnodeProposeWriteMsg
000000000037dd60 T _vnodeQueryClose
000000000037dcf0 T _vnodeQueryOpen
000000000037dd30 T _vnodeQueryPreClose
00000000003856a0 T _vnodeRedirectRpcMsg
000000000037f320 T _vnodeResetLoad
000000000037d440 T _vnodeRollback
000000000037c3a0 T _vnodeSaveInfo
00000000003800e0 T _vnodeScheduleTask
000000000037c200 T _vnodeShouldCommit
000000000037c320 T _vnodeShouldCommitOld
000000000037d350 T _vnodeShouldRollback
0000000000389ee0 T _vnodeSnapRead
0000000000389e20 T _vnodeSnapReaderClose
0000000000389cd0 T _vnodeSnapReaderOpen
000000000038aee0 T _vnodeSnapWrite
000000000038aa80 T _vnodeSnapWriterClose
000000000038a8f0 T _vnodeSnapWriterOpen
00000000003793b0 T _vnodeStart
00000000003793d0 T _vnodeStop
00000000003885f0 T _vnodeSyncCheckTimeout
0000000000388580 T _vnodeSyncClose
000000000037d310 T _vnodeSyncCommit
0000000000387c90 T _vnodeSyncOpen
0000000000388510 T _vnodeSyncPostClose
0000000000388400 T _vnodeSyncPreClose
0000000000388320 T _vnodeSyncStart
000000000037c180 T _vnodeUpdCommitSched
00000000003855e0 T _vnodeUpdateMetaRsp
000000000037bde0 T _vnodeValidateTableHash
000000000037fcc0 T _vnode_done_commit
000000000037fca0 T _vnode_wait_commit
0000000000441c70 T _voteGrantedCreate
0000000000441d70 T _voteGrantedDestroy
[...]

PR #19855, opened 2023-02-07
Fix unnecessary exported functions

freemine:
https://github.com/taosdata/TDengine/issues/19665

has been successfully tested via https://github.com/freemine/taos_odbc locally on my machine, for linux/macos/windows(11).
i did not thoroughly check if test-cases in TDengine can pass or not, but i strongly believe they shall pass if they do not rely on those `internal-but-exported` functions.
for other client-side-tools, such as taosadapter/all kinds of taos-connectors, have no time to test one by one, but they shall pass just as `taos` shell tool does.

### DIFF
--- a/include/client/taos.h
+++ b/include/client/taos.h
@@ -94,7 +94,7 @@ typedef struct TAOS_FIELD_E {
   int32_t bytes;
 } TAOS_FIELD_E;
 
-#ifdef WINDOWS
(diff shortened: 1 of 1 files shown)


freemine:
to thoroughly handle this issue, it'll take much more consideration in TDengine's cmake-build-system.
thus, this is PR is closed.

PR #19876, opened 2023-02-08
make_install.bat: (#19658)

facetosea:
1. copy files failed if svc-deleting is still outstanding
2. unnecessary duplication of C:\TDengine in %PATH%


sangshuduo (approved):
LGTM

PR #19894, opened 2023-02-09
fix: taos shell show err msg twice with cloud service

sangshuduo:


sangshuduo:
already reviewed in #19892

PR #19911, opened 2023-02-10
fix: instructions for starting the service on the mac

facetosea:


sangshuduo (changes requested):
Need hold until the best way to launch daemon/agents reviewed.

sangshuduo:
replaced with #19963

PR #20045, opened 2023-02-18
fix: support meta sync between mnode and vnode

kailixu:


kailixu:
Internal version, no need to merge to main branch.

PR #20047, opened 2023-02-19
enh: check if the directory exists before installing on mac

facetosea:


packaging/tools/post.sh
```diff
@@ -627,9 +627,16 @@ function install_app() {
   fi
 }
 
-function install_TDengine() {
-    echo -e "${GREEN}Start to install TDengine...${NC}"
-    log_print "start to install TDengine"
+function checkDirectory() {
+    if [ ! -d "/usr/local/bin" ]; then
+      sudo mkdir -p /usr/local/bin
```
sangshuduo: need consider an environment like docker has no sudo command.

sangshuduo (changes requested):
see my comments

packaging/tools/post.sh
facetosea: Adapted to different platforms

sangshuduo (approved):
LGTM

sangshuduo (approved):
LGTM

PR #20156, opened 2023-02-24
fix(query): fix apply limit error for doMultiwayMerge operator if datablocks belongs to the same group

glzhao89:
datablocks within same group


dapan1121 (approved):

feici02:
Fix #20281.

PR #20195, opened 2023-02-27
feat: include libuv errno into taos-errno

freemine:
when uv_loop_init failed, such as, 'Too many files opened', set `terrno` with those from `libuv`.


freemine:
seems ulimit-issue has been solved,  test in `3bd66ecb52ade6ca874fcc1628c961773636e5dd`

PR #20232, opened 2023-03-01
fix: windows release from tdengine/package/release.bat failed

facetosea:


sangshuduo (approved):

sangshuduo (changes requested):
use "cusName" "cusPrompt" to align with other platform. use "TDengine" if it will show as brand name.

sangshuduo (approved):
LGTM

PR #20252, opened 2023-03-03
enh: fsync WAL idx and log when walRollImpl

bgzhao66:


xiao-yu-wang:
7-tmq/tmqConsFromTsdb1-mutilVg-mutilCtb.py failed because of the CI itself. It has nothing to do with this commit. It can be merged.

PR #20331, opened 2023-03-08
docs: this line is inaccurate, in current connector-python repo, it r…

AdamEECS:
…equire python >= 3.6.2 because there is a dependence iso8601


sangshuduo (approved):
LGTM

PR #20431, opened 2023-03-13
test: update-run_case.sh

AdamEECS:


AdamEECS (commented):
with func typing, try to figure out error.

PR #20432, opened 2023-03-13
test: update run_case.sh with branch fix/adamji/remove-typing

AdamEECS:


AdamEECS (commented):
removed typing in func def line

PR #20486, opened 2023-03-15
Do not retrieve err msg when connection is established successfully 

robotspace:
Fix bug that triggers an segment exception and add g flag in build command for GDB debug convinience.


sangshuduo (approved):
LGTM

PR #20751, opened 2023-04-04
feat: support split vgroup

bgzhao66:


bgzhao66:
Merge to 3.0 instead of main. Close it.

PR #20832, opened 2023-04-10
fix: handle leak

facetosea:


sangshuduo (approved):
LGTM. Best practice may be just delete unused code

facetosea:
Has been fixed.

PR #20929, opened 2023-04-14
docs(taosAdapter): the automatic database creation behavior for the schemaless protocol can be configured

huskar-t:


sangshuduo (approved):
LGTM

PR #21027, opened 2023-04-23
docs: add badge for  python connector

AdamEECS:


sangshuduo (changes requested):
My suggestion does not to add any badge in the official document to avoid extra step to exclude if someone need to generate the user manual based on the document markdown. @zitsen

PR #21037, opened 2023-04-23
Feature/3.0 geometry

dinglezhang:
This pull request is a feature to add Geometry support into TDengine, just like what PostGIS does on PostgreSQL and what SpatiaLite does on SQLlite.
They are the followings:

1. Add Geometry type

2. Add Geometry IO functions:
    ST_GeomFromText()
    ST_AsText()

3. Add Geometry relation functions:
    ST_Intersects()
    ST_Equals()
    ST_Touches()
    ST_Covers()
    ST_Contains()
    ST_ContainsProperly()

All of them comply to OGC defined functions. I have added docs about the above, but only in English version so far.

The implementations mainly include the followings:

1. Add TSDB_DATA_TYPE_GEOMETRY data type.
    It looks like TSDB_DATA_TYPE_BINARY, but is used for Geometry function specially. So I had to add logic like 'if (data.type == TSDB_DATA_TYPE_GEOMETRY)' along with 'if (data.type ==TSDB_DATA_TYPE_BINARY)', that is why my changes cover so many files.
    I have added GEOMETRY syntax into source/libs/parser/inc/sql.y, and generated new source/libs/parser/src/sql.c by calling lemon.
    So far I use WKB (OGC well-known binary) as internal storage format for all kinds of geometries, including POINT, LINESTRING, POLYGON, MULTIPOINT, MULTILINESTRING, MULTIPOLYGON, GEOMETRYCOLLECTION.

2. Add a new lib source/libs/geometry to wrap functions provided by GEOS (https://libgeos.org/doxygen/geos__c_8h.html) introduced, and implement geometry functions for SQL.
    And register these functions in source/libs/function/src/builtins.c.

3. Add parseGeometry() in source/libs/parser/src/parInsertSql.c, so that we can parse WKT (OGC well-known text) automatically, just like we can parse time by parseTime() for user input time text.
    Add shellPrintGeometry() in tools/shell/src/shellEngine.c, so that we can get WKT results for geometry type in taos client.

4. Add unit test (source/libs/geometry/test) for geometry lib.
    Add system test (tests/system-test/2-query/geometry.py) for geometry type and SQL functions.

5. Add docs about geometry type and functions, but only English version.

6. Changes necessary on cmake (especially to add geometry lib dependency internally and GEOS dependency externally) and README.md.

7. Built on ubuntu and Windows. Unit test passed, and most system test passed.

I really need the feature when I wanted to use TDengine in our project. So I tried to do the feature in several months. I suppose it is need by IoT applications since most things have their positions to store and need to do some geometry calculations on db server side.

It is my honor to contribute to TDengine, one of the most popular time serial databases. I am looking forward to the acceptance of my changes.
I know the feature is a bit big change. It may be better to create a new feature branch instead of 3.0. I would like to merge to the new feature branch.

The current implementation is usable, but only with very basic functionality. I plan to implement more things around geometry, after my changes are merged.

### DIFF
--- a/cmake/cmake.platform
+++ b/cmake/cmake.platform
@@ -57,6 +57,8 @@ IF (${CMAKE_SYSTEM_NAME} MATCHES "Linux" OR ${CMAKE_SYSTEM_NAME} MATCHES "Darwin
         SET(TD_DARWIN TRUE)
         SET(OSTYPE "macOS")
         ADD_DEFINITIONS("-DDARWIN -Wno-tautological-pointer-compare")
+        INCLUDE_DIRECTORIES(/usr/local/include)
+        LINK_DIRECTORIES(/usr/local/lib)
 
         IF (${CMAKE_SYSTEM_PROCESSOR} MATCHES "arm64")
             MESSAGE("Current system arch is arm64")
--- a/source/util/CMakeLists.txt
+++ b/source/util/CMakeLists.txt
@@ -5,9 +5,6 @@ if (DEFINED GRANT_CFG_INCLUDE_DIR)
   add_definitions(-DGRANTS_CFG)
 endif()
 
-if(NOT TD_WINDOWS)
-    INCLUDE_DIRECTORIES(/usr/local/include)
-endif(NOT TD_WINDOWS)
 target_include_directories(
     util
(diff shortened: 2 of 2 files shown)


zitsen:
Awesome!

dingbo8128:
Looks great!

PR #21040, opened 2023-04-23
docs: add docs for `taoskeeper with prometheus`

sunpe:


huolibo (approved):
LGTM

sangshuduo (changes requested):
suggest adding the new content to the taoskeeper section instead of Prometheus' as the purpose is to explain how taoskeeper to send the metrics to Prometheus

PR #21061, opened 2023-04-24
fix unnecessary mkdir failure

freemine:
```
failed to create dir:/var/log/taos/ since Operation not permitted WARING: Create taoslog failed:Operation not permitted. configDir=/etc/taos/
```

### DIFF
--- a/source/os/src/osDir.c
+++ b/source/os/src/osDir.c
@@ -206,7 +206,15 @@ int32_t taosMulModeMkDir(const char *dirname, int mode) {
 #endif
 
   if (taosDirExist(temp)) {
-    return chmod(temp, mode);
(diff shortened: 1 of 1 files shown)


sangshuduo (approved):

sangshuduo (commented):
Wonder if this PR will conceal the permission issue that should report to the upper layer.  Could we fix it in taos shell?

freemine:
> Wonder if this PR will conceal the permission issue that should report to the upper layer.  Could we fix it in taos shell?

Or it might be the issue that whether taos.log/taosd.log in the same directory?

WeiduWedo:
我也遇到同样的问题

PR #21223, opened 2023-05-09
docs(driver): jdbc 3.2.1 document and error code

huolibo:


gccgdb1234:
Please also change english version, you can **simply copy same changes of ZH version to EN, and then submit**, I will pull and change  the EN part.

huolibo:
> Please also change english version, you can **simply copy same changes of ZH version to EN, and then submit**, I will pull and change the EN part.

I have already modified the EN version. Is there anything else that needs to be changed?

PR #21484, opened 2023-05-26
enhance: python udf sample with numpy

shenglian-zhou:


cdiwadkar16 (approved):
This is a good enhancement. I think showing that an external library that is not part of the standard Python library is helpful.

PR #21567, opened 2023-06-02
fix: prevent projectoperator scanning all rows when limit with no group

wangjiaming0909:


source/libs/executor/src/groupoperator.c
```diff
@@ -647,6 +647,8 @@ uint64_t calcGroupId(char* pData, int32_t len) {
   // NOTE: only extract the initial 8 bytes of the final MD5 digest
   uint64_t id = 0;
   memcpy(&id, context.digest, sizeof(uint64_t));
+  if (0 == id)
+    memcpy(&id, context.digest + 8, sizeof(uint64_t));
```
hjxilinx, line 651: Is this necessary? 
Basically, it is impossible for an MD5 digest in the case that the initial 8 bytes is 0, but the following 8 bytes are non-zero value.

dapan1121 (approved):

PR #21608, opened 2023-06-06
fix(cache/binary): fix crash in freeItem

stephenkgu:


gccgdb1234 (approved):

gccgdb1234:
./test.sh -f tsim/stream/distributeInterval0.sim
this test case is not stable

PR #21762, opened 2023-06-16
fix: show queries returned less rows when one conn has more than 100 queries

wangjiaming0909:


source/dnode/mnode/impl/src/mndProfile.c
```diff
@@ -764,104 +764,129 @@ static int32_t mndRetrieveConns(SRpcMsg *pReq, SShowObj *pShow, SSDataBlock *pBl
...
 
+  while (numOfRows < rows) {
+    pConn = mndGetNextConn(pMnode, pShow->pIter);
+    if (pConn == NULL) {
+      pShow->pIter = NULL;
+      break;
+    }
+
+    int32_t packedRows = packQueriesIntoBlock(pShow, pConn, pBlock, 0, rows - numOfRows);
+    pShow->curIterPackedRows = packedRows;
```
dapan1121, line 887: +=

PR #21771, opened 2023-06-19
update test case to combine three cases into one, reduce the time cos…

flyingangel2013:
…t of inserting data by charles


tomchon:
LGTM

PR #22062, opened 2023-07-13
Correct some typos in string literals

sunsided:
This is a small PR that corrects some typos in string literals, such as `initizlized` instead of `initialized`, etc.


sangshuduo:
@sunsided thanks for your contribution!

PR #22189, opened 2023-07-25
enh: change default WAL_RETENTION_PERIOD as 3600

bgzhao66:


bgzhao66:
Merge to 3.0 only. This PR is to be closed.

PR #22510, opened 2023-08-21
update packaging script

xiaopingcs:
1. remove community package testing from ubuntu
2. add tdef.h into release package


sangshuduo (approved):
LGTM

PR #22533, opened 2023-08-23
feat: add minDiskFreeSize cfg

hzcheng:


gccgdb1234:
There is one test case that fails already in the 3.0 branch, so merge this PR forcefully

PR #22858, opened 2023-09-12
fix read line on windows

facetosea:


source/os/src/osFile.c
```diff
@@ -841,14 +841,33 @@ int64_t taosGetLineFile(TdFilePtr pFile, char **__restrict ptrBuf) {
...
+    bytesRead = strlen(*ptrBuf + totalBytesRead);
+    totalBytesRead += bytesRead;
+
+    if (totalBytesRead < bufferSize - 1 || (*ptrBuf)[totalBytesRead - 1] == '\n') {
+      break;
+    }
+
+    bufferSize += 512;
+    *ptrBuf = taosMemoryRealloc(*ptrBuf, bufferSize);
+    if (*ptrBuf == NULL) return -1;
```
kailixu, line 866: the ptrBuf should be released

PR #22897, opened 2023-09-14
merge stable sort intead of qsort

facetosea:


facetosea:
I need to merge these commits.

PR #23120, opened 2023-09-29
fix: reset hb flag for duplicated user key

kailixu:


kailixu:
close since too much unexpected changes

PR #23128, opened 2023-10-02
fix: support check/merge active code

kailixu:


kailixu:
close since main and 3.0 differ too much

PR #23132, opened 2023-10-04
fix: mutex init for new trans

kailixu:


kailixu:
close since another method used and merged.

PR #23411, opened 2023-10-25
enh: check interval and keep for retentions

kailixu:


kailixu:
already merged with other PR

PR #23427, opened 2023-10-27
fix: interval more than 1000 years

facetosea:


source/libs/parser/src/parTranslater.c
```diff
@@ -3442,6 +3456,7 @@ static void convertVarDuration(SValueNode* pOffset, uint8_t precision) {
   pOffset->unit = units[precision];
 }
 
+static const int64_t  maxKeepMS = (int64_t)3600 * 1000 * 24 * (365 * 1000 + 250);
```
dapan1121: 用KEEP最大值的宏

facetosea: done

dapan1121 (approved):

PR #23782, opened 2023-11-23
enh: not allow to acquire a vnode in failed mode from vmAcquireVnode

bgzhao66:


bgzhao66:
Since it is a wrong JIRA ID, close this PR.

PR #23843, opened 2023-11-28
fix crash when insert big csv

bobliubj:
The big csv file will be split into multiple imports, and the crash is caused by parameter configuration problems during subsequent batch imports. the solution is refer the previous batch configuration

### DIFF
--- a/source/libs/parser/src/parInsertSql.c
+++ b/source/libs/parser/src/parInsertSql.c
@@ -1584,7 +1584,7 @@ typedef struct SStbRowsDataContext {
   bool           isJsonTag;
 } SStbRowsDataContext;
 
-typedef struct SRowsDataContext{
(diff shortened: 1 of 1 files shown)


source/libs/parser/src/parInsertSql.c
```diff
@@ -1819,6 +1824,10 @@ static int32_t parseOneStbRow(SInsertParseContext* pCxt, SVnodeModifyOpStmt* pSt
     if (TSDB_CODE_SUCCESS == code) {
       insCheckTableDataOrder(pTableDataCxt, TD_ROW_KEY(*pRow));
     }
+
+    void* pData = pTableDataCxt;
```
dapan1121: 这是SQL和CSV都会经过的流程

dapan1121 (approved):

PR #23927, opened 2023-12-04
enh : accept float data as int field input

bobliubj:
For integer fields, users are allowed to pass in floating point numbers, hexadecimal numbers, and binary numbers.
For floating-point numbers, rounding is used as the field result value.
If it is converted to an integer and exceeds the field expression range, an error is reported

### DIFF
--- a/tests/system-test/1-insert/insert_double.py
+++ b/tests/system-test/1-insert/insert_double.py
@@ -40,8 +40,8 @@ class TDTestCase:
 
         # str support
         tdSql.execute(f"insert into {table_name} values(now, '-16', '+6')")
-        tdSql.execute(f"insert into {table_name} values(now, ' -80.99 ', '  +0042  ')")
-        tdSql.execute(f"insert into {table_name} values(now, ' -0042  ', ' +80.99 ')")
+        tdSql.execute(f"insert into {table_name} values(now, ' -80.99', ' +0042')")
+        tdSql.execute(f"insert into {table_name} values(now, ' -0042', ' +80.99')")
         tdSql.execute(f"insert into {table_name} values(now, '52.34354', '18.6')")
         tdSql.execute(f"insert into {table_name} values(now, '-12.', '+5.')")
         tdSql.execute(f"insert into {table_name} values(now, '-.12', '+.5')")


source/common/src/tvariant.c
```diff
@@ -19,17 +19,215 @@
...
+
+    if (z[1] == 'x' || z[1] == 'X') {
+      // parsing as hex
+      return parseHexUInteger(z, n, value);
+    }
+  }
+
+  // rm flag u-unsigned, l-long, f-float(if not in hex str)
+  char last = tolower(z[n - 1]);
+  if (last == 'u' || last == 'l' || last == 'f') {
```
dapan1121: 支持这几个字符的意义是什么？

source/common/src/tvariant.c
```diff
@@ -19,17 +19,215 @@
...
+
+int32_t parseSignAndUInteger(const char *z, int32_t n, bool *is_neg, uint64_t *value) {
+  *is_neg = false;
+  if (n < 1) {
+    return TSDB_CODE_FAILED;
+  }
+
+  // parse sign
+  bool has_sign = false;
+  if (z[0] == '-') {
```
dapan1121, line 51: 浮点数写入支持吗

source/common/src/tvariant.c
```diff
@@ -19,17 +19,215 @@
...
+    return TSDB_CODE_FAILED;
+  }
+  *value = round(val);
+  return TSDB_CODE_SUCCESS;
+}
+
+int32_t removeSpace(const char **pp, int32_t n) {
+  // rm blank space from both head and tail
+  const char *z = *pp;
+  while (*z == ' ' && n > 0) {
```
dapan1121: 其他字符比如换行、TAB呢

tests/system-test/1-insert/insert_double.py
```diff
@@ -0,0 +1,132 @@
...
+        self.db = "db1"
+        tdLog.debug(f"start to excute {__file__}")
+        tdSql.init(conn.cursor(), False)
+
+    def prepare_db(self):
+        tdSql.execute(f"drop database if exists {self.db}")
+        tdSql.execute(f"create database {self.db}")
+        tdSql.execute(f"use {self.db}")
+
+    def test_value(self, table_name, dtype, bits):
```
dapan1121, line 25: 缺少异常用例

dapan1121 (commented):
1. 检查哪些规则有必要，是否需要应用到其他类型的写入，比如+、-、空格那浮点数写入是不是也要支持？
2. 测试一下正常写入的性能影响，比如10列整数写入场景

source/common/src/tvariant.c
bobliubj: 支持

source/common/src/tvariant.c
bobliubj: 尾部space若不清理，写入时提示字符串非法。其他字符过滤暂不支持

PR #24010, opened 2023-12-11
feat: compact monitor and kill

hzcheng:


source/common/src/tmsg.c
```diff
@@ -4276,6 +4337,7 @@ int32_t tDeserializeSRetrieveTableReq(void *buf, int32_t bufLen, SRetrieveTableR
   if (tDecodeCStrTo(&decoder, pReq->tb) < 0) return -1;
   if (tDecodeCStrTo(&decoder, pReq->filterTb) < 0) return -1;
   if (tDecodeCStrTo(&decoder, pReq->user) < 0) return -1;
+  if (tDecodeI64(&decoder, &pReq->compactId) < 0) return -1;
```
hzcheng: Please assure backward compactibility

shenglian-zhou: done

PR #24316, opened 2024-01-02
test: add testcase that merge same ver stt file after split db

tomchon:


tests/system-test/0-others/splitVGroupRep1.py
```diff
@@ -412,8 +412,16 @@ def compactAndCheck(self):
     def run(self):
         # prepare env
         self.prepareEnv()
-
+        tdLog.info("add  two  same ver stt file like v4f1944ver3.stt and v4f1944ver4.stt to db2 and db1 ")
```
bgzhao66: 这个说明，可以调整一下：generate at least two stt files of the same fileset (e.g. v4f1944) for both db1 and db2.

bgzhao66 (commented):
Post comments.

DuanKuanJun (approved):

Issue #24416, opened 2024-01-10
gcc compile error


robotspace:
**Bug Description**
gcc compile error

**To Reproduce**
1. Go to source path
2. cmake .. -DBUILD_TOOLS=false -DBUILD_CONTRIB=false
3. See error /home/src/tdengine/contrib/libuv/src/queue.h:68:19: error: storing the address of local variable ‘queue’ in ‘((void * (**)[2])MEM[(void *[2] * *)loop_13(D) + 16B])[1]’ [-Werror=dangling-pointer=]
   68 |     QUEUE_PREV(q) = (n);                                                      
                         ^
**Expected Behavior**
compile successfully

**Screenshots**
none

**Environment (please complete the following information):**
 - OS: fedora38
 - gcc version: gcc version 13.2.1 20231011
 - TDengine Version : current version on master branch

**Additional Context**
none

robotspace:
Warning dangling-pointer is imported since gcc13.0 . Many libs implemented by C will meet this problem. Build env should be set over gcc13 for test purpose.

PR #24451, opened 2024-01-12
Remove compile error of dangling pointer caused by libuv module in gcc13 environment.

robotspace:
Warning dangling-pointer is imported since gcc13.0. Tdengine's configure retreat it as error: https://github.com/taosdata/TDengine/issues/24416
 Libuv has fix this false positive report on latest version: https://github.com/libuv/libuv/pull/4022


robotspace:
Here is an duplicate PR. https://github.com/taosdata/TDengine/pull/23608

PR #24583, opened 2024-01-22
Case/td 28027 3.0 new add srvCtl cluster support

menshibin:


tests/army/frame/server/dnodes.py
```diff
@@ -47,6 +47,7 @@ def __init__(self):
         self.valgrind = 0
         self.asan = False
         self.killValgrind = 0
+        self.model = "dnode"
```
DuanKuanJun: 修改成 single

DuanKuanJun (approved):

PR #24809, opened 2024-02-21
enh: ensure redos and undos of parallel trans being of the same type

bgzhao66:


bgzhao66:
Ensure that the redos and undos of TRN_EXEC_PARALLEL mnd trans are of the same type.

hzcheng (approved):

PR #25148, opened 2024-03-21
fix: unify tag value parsing syntax

kailixu:


kailixu:
https://jira.taosdata.com:18080/browse/TS-4421
https://jira.taosdata.com:18080/browse/TD-29209

PR #25238, opened 2024-04-01
fix: build error on loongarch64

KFERMercer:
still missing:
 - libdmodule.a
 - librocksdb.a


KFERMercer:
please add the 2 binaries mentioned above to complete this commit, think you

PR #25419, opened 2024-04-19
Fix colCmpr initialization in metaDecodeEntry

yihaoDeng:
This pull request fixes the initialization of colCmpr in the metaDecodeEntry function. Previously, the initialization was missing in certain cases, leading to potential issues. This PR ensures that colCmpr is properly initialized in all cases.


yihaoDeng (commented):
test

PR #26149, opened 2024-06-14
docs: Remove unnecessary geos-devel dependency in Ubuntu installation

bitcapybara:
When compiling the TDengine source code, the geo-related dependencies will be automatically downloaded. Additionally, there is no geos-devel dependency in the Ubuntu/Debian dependency repository. This dependency needs to be removed from the README, otherwise an error will occur when executing the dependency installation statement.

image


tomchon:
it's should be "libgeos-dev " @bitcapybara ,i will modify it ,thanks

PR #26438, opened 2024-07-05
fix(TD-30611-3.0): run CI test for taos_tools

xinjiempolde:


xinjiempolde:
The CI tests about taos_tools have passed, close this PR

PR #26960, opened 2024-08-02
fix(stream): check status before start init, do some internal refactor.

hjxilinx:


hjxilinx:
delete-check.py failed.

PR #27220, opened 2024-08-14
tetst: using clang-query-16 instead of 10 when testing return values

tomchon:


tomchon:
https://github.com/taosdata/TDengine/pull/27170

PR #27530, opened 2024-08-29
enh: (s3) continue to check other s3 cfg after an error occurs

LiShunGang:


tomchon:
@LiShunGang  得重新提交才行，或者提交空的。不然 pr 没法重跑了

PR #28005, opened 2024-09-20
fix(query)[TD-32196]. Fix error code handling in group sort

JinqingKuang:
Resolved an issue in doGroupSort where a failed function call did
not properly handle the error code. The error code is now propagated
upwards to ensure proper error reporting and handling in higher layers.


jtao1735:
When you submit the code, please write some description about your changes

dapan1121 (approved):

PR #28162, opened 2024-09-27
test: modify crash_gen default taos.cfg

jiajingbin:


flyingangel2013 (approved):
update taos.cfg for crash0-gen tool， doesn't need CI results

PR #28175, opened 2024-09-29
docs(kafka): add missing metrics for kafka datain

zitsen:
Close [TD-31388](https://jira.taosdata.com:18080/browse/TD-32388)


feici02 (approved):
LGTM

PR #28249, opened 2024-10-08
enh(query)[TS-4661] Add AVX2 support for decompression and MIN/MAX comparison

JinqingKuang:
- Implement AVX2 instructions to decompress float and double data types
- Add AVX2-based MIN/MAX comparison for both integer and floating-point data types
- Include unit tests to verify functionality and performance improvements


cmake/cmake.define
```diff
@@ -172,21 +172,19 @@ ELSE ()
         SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -msse4.2")
     ENDIF()
 
-    IF ("${SIMD_SUPPORT}" MATCHES "true")
```
hjxilinx: should check the simd_support macro to avoid generating the implicit fma instructions by compiler.

source/util/src/tcompression.c
```diff
@@ -467,70 +471,99 @@ int32_t tsDecompressINTImp(const char *const input, const int32_t nelements, cha
     return nelements * word_length;
   }
 
-#if __AVX2__
-  tsDecompressIntImpl_Hw(input, nelements, output, type);
```
hjxilinx, line 471: Are there some problems with this function?

include/util/tcompression.h
```diff
@@ -152,11 +152,15 @@ int32_t tsDecompressBigint(void *pIn, int32_t nIn, int32_t nEle, void *pOut, int
...
 int32_t tsDecompressIntImpl_Hw(const char *const input, const int32_t nelements, char *const output, const char type);
-void    tsDecompressFloatImplAvx512(const char *const input, const int32_t nelements, char *const output);
-void    tsDecompressFloatImplAvx2(const char *const input, const int32_t nelements, char *const output);
+int32_t tsDecompressFloatImpAvx2(const char *input, int32_t nelements, char *output);
+int32_t tsDecompressDoubleImpAvx2(const char *input, int32_t nelements, char *output);
+void    tsDecompressTimestampAvx2(const char *input, int32_t nelements, char *output, bool bigEndian);
+#endif
+#ifdef __AVX512VL__
 void tsDecompressTimestampAvx512(const char *const input, const int32_t nelements, char *const output, bool bigEndian);
-void tsDecompressTimestampAvx2(const char *const input, const int32_t nelements, char *const output, bool bigEndian);
```
hjxilinx, line 159: And this one

cmake/cmake.define
JinqingKuang: Resolved.

source/util/src/tcompression.c
JinqingKuang: No, it's just that its performance is poor, so it wasn't used.

include/util/tcompression.h
JinqingKuang: Same as above.

PR #28323, opened 2024-10-11
update tag

xiaopingcs:


flyingangel2013 (approved):
only update make script， doesn't need CI running results

PR #28755, opened 2024-11-13
test:delete useless file

tomchon:


feici02 (approved):
delete script which should be deleted in #28751

feici02 (approved):

PR #28789, opened 2024-11-15
docs(opc): node_id_pattern and browse_name_pattern#TD-31908

zyyang90:
docs(opc): node_id_pattern and browse_name_pattern#TD-31908


pigzhou (approved):
LGTM

PR #28923, opened 2024-11-25
ci: modifying the way of writing the matching strings for the grep

tomchon:
# Pull Request Checklist


feici02 (approved):
LGTM

PR #29372, opened 2024-12-27
fix ps/grep/kill flaws found in test.script

freemine:
# Description

when <TDengine> source tree is stored in the path which contains `taos`, such as $HOME/taos/TDengine
./test.sh -f ...sim will fail in $HOME/taos/TDengine/tests/script

# Checklist
`pushd tests/script && ./test.sh -f ./tsim/join/join.sim; popd`


dapan1121 (commented):
LGTM

guanshengliang (approved):

Issue #29766, opened 2025-02-13
seems unnecessary pcre2 used in geos functionality


freemine:
**Performance Issue** 
_it seems doRegComp/doRegExec in geosWrapper.c does not have any effect except loss of both performance and accurate parse error message to demonstrate._

**Problem Description** 
_WKT input string finally gets parsed in GEOSWKTReader_read_r, and error message is stored because of previous config by_
```
(void)GEOSContext_setErrorMessageHandler_r(geosCtx->handle, geosErrMsgeHandler, geosCtx->errMsg); 
```
_thus, we don't see any better effect introduced by_
```
if (doRegExec(inputWKT, geosCtx->WKTRegex, geosCtx->WKTMatchData) != 0) {
  code = TSDB_CODE_FUNC_FUNTION_PARA_VALUE;
  goto _exit;
}
```

**To Reproduce**
_None._

**Screenshots**
_original code snippet and negative case:_
<img width="960" alt="Image" src="https://github.com/user-attachments/assets/8dd7807e-40f3-4998-a4a7-caba462b6c0a" />
<img width="1018" alt="Image" src="https://github.com/user-attachments/assets/cf471db5-3476-40d2-a536-3d79cfb23ad4" />

_compared to modified version:_
<img width="984" alt="Image" src="https://github.com/user-attachments/assets/f46f8993-7906-42a8-b36a-e4726a7c8ecd" />
<img width="1371" alt="Image" src="https://github.com/user-attachments/assets/0b4db48f-31c3-4cdc-9ff2-188681859247" />

**Environment (please complete the following information):**
* OS: [ubuntu 20.04]
* Memory, CPU, current Disk Space
* TDengine Version [git commit: d4dddb5fd7f86528614ae89e1fab313ae8ab923e]

**Additional Context** 
_None._

Pengrongkun:
doRegExec check geometry keyword error report

examples：
insert into t_ge values(now, ' POINTT(1.0 2.0)');
insert into t_ge values(now, 'LINESTRINGG(1.00 +2.0, -3.2 3.0, 5.00 5.01, 5.0 6.0)');
insert into t_ge values(now, 'POLYGONNNN((1.0 1.0, -2.0 +2.0, 1.0 1.0))');

PR #29768, opened 2025-02-13
remove tpcre2.h and related which seems unnecessary

freemine:
# Description
#29766
because **_pcre2_** seems only used in such case, pcre2_CMakeLists.txt.in and other related are also removed by this PR.

# Checklist


freemine:
as detailed in #29766
strict mode is expected in original version, but this PR aligns with libgeos which is not strict.
eg.: POINTabc (1 2) is not valid originally, while libgeos will accept it as if it's POINT (1 2)
though libgeos could be modified to make it strict, such as in src/io/WKTReader.c
```
bool
WKTReader::isTypeName(const std::string & type, const std::string & typeName) {
  if (type == typeName) return true;
  if (type == (typeName + "Z")) return true;
  if (type == (typeName + "M")) return true;
  if (type == (typeName + "ZM")) return true;
  return false;
}
```
and some other places.

PR #29906, opened 2025-02-23
feat(query)[TS-5470]: add syntax to specify minimum duration for event and state windows

JinqingKuang:
# Description

Introduce the `true for` syntax to allow users to specify the minimum duration for event and state windows. Add corresponding tests to validate the feature. Updated the user manual with usage instructions.

# Checklist

- [x] Is the user manual updated?


docs/en/14-reference/03-taos-sql/12-distinguished.md
```diff
@@ -177,6 +177,12 @@ TDengine also supports using CASE expressions in state quantities, which can exp
 SELECT tbname, _wstart, CASE WHEN voltage >= 205 and voltage <= 235 THEN 1 ELSE 0 END status FROM meters PARTITION BY tbname STATE_WINDOW(CASE WHEN voltage >= 205 and voltage <= 235 THEN 1 ELSE 0 END);
 ```
 
+The status window supports using the TRUE_FOR parameter to set its minimum duration. If the window's duration is less than the specified value, it will be discarded automatically and no result will be returned. For example, setting the minimum duration to 3 seconds:
```
dapan1121: state window

PR #30014, opened 2025-03-05
Tini does not effectively monitor the running status of the child process here. I made some changes

hchw:
# Description

Tini does not effectively monitor the running status of the child process here. I made some changes, I understand that tini's posture is not right. Not sure if the author has other intentions

# Checklist

- [*] Is the user manual updated?

### DIFF
--- a/packaging/docker/bin/entrypoint.sh
+++ b/packaging/docker/bin/entrypoint.sh
@@ -74,8 +74,16 @@ fi
 
 PID_TAOSADAPTER=0
 if [ "$DISABLE_ADAPTER" = "0" ]; then
-    which taosadapter >/dev/null && taosadapter &
-    PID_TAOSADAPTER=$!
+    # check if the process is running
+    if ! pgrep taosadapter >/dev/null; then
+        which taosadapter >/dev/null && {
+            taosadapter &
+            PID_TAOSADAPTER=$!
+        }
+    else
+        PID_TAOSADAPTER=$(pgrep taosadapter)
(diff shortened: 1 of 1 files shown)


packaging/docker/bin/entrypoint.sh
```diff
@@ -66,21 +68,41 @@ else
...
 
+PID_TAOSADAPTER=0
 if [ "$DISABLE_ADAPTER" = "0" ]; then
-    which taosadapter >/dev/null && taosadapter &
+    # check if the process is running
+    if ! pgrep taosadapter >/dev/null; then
+        which taosadapter >/dev/null && {
+            taosadapter &
+            PID_TAOSADAPTER=$!
+        }
```
jiajingbin, line 82: Suggest extracting a common method for taosadapter,taoskeeper and taos-explorer, such as:

`start_process() {
    local process_name=$1
    local port=$2
    local disable_flag=$3
    local pid_var=$4

    if [ "$disable_flag" = "0" ]; then
        # Check if the process is already running
        if ! pgrep "$process_name" >/dev/null; then
            which "$process_name" >/dev/null && {
                "$process_name" &
                eval "$pid_var=$!"
            }
        else
            eval "$pid_var=$(pgrep "$process_name")"
            echo "$process_name already running with PID: ${!pid_var}"
        fi

        # Wait for the port to be ready
        for _ in $(seq 1 20); do
            nc -z localhost "$port" && break
            sleep 0.5
        done
    fi
}`

PR #33063, opened 2025-09-24
feat: rollup sma

kailixu:
# Description

Please briefly describe the code changes in this pull request.

# Checklist

### DIFF
--- a/source/dnode/vnode/src/tsdb/tsdbRetention.c
+++ b/source/dnode/vnode/src/tsdb/tsdbRetention.c
@@ -434,7 +434,7 @@ static bool tsdbInRetentionTimeRange(STsdb *tsdb, int32_t fid, STimeWindow tw, i
     TSKEY  minKey, maxKey;
     int8_t precision = tsdb->keepCfg.precision;
     if (precision < TSDB_TIME_PRECISION_MILLI || precision > TSDB_TIME_PRECISION_NANO) {
-      tsdbError("vgId:%d, invalid precision %" PRIi8, TD_VID(tsdb->pVnode), precision);
+      tsdbError("vgId:%d, failed to check retention time range since invalid precision %" PRIi8, TD_VID(tsdb->pVnode), precision);
       return false;
     }
     tsdbFidKeyRange(fid, tsdb->keepCfg.days, precision, &minKey, &maxKey);


include/common/tmsg.h
```diff
@@ -199,9 +199,23 @@ typedef enum _mgmt_table {
...
+  TSDB_MGMT_TABLE_RETENTION,
+  TSDB_MGMT_TABLE_RETENTION_DETAIL,
   TSDB_MGMT_TABLE_MAX,
 } EShowType;
 
+typedef enum {
+  TSDB_OPTR_NORMAL = 0,  // default
+  TSDB_OPTR_SSMIGRATE = 1,
+  TSDB_OPTR_ROLLUP = 2,
+} ETsdbOpType;
```
localvar, line 212: if these names are only for compact,  then they are too general to show their usage well, and make related code hard to read in long time maintaintence. propose to use something like `ETsdbCompactReason`, `TSDB_COMPACT_REASON_NORMAL`.

source/dnode/vnode/src/tsdb/tsdbRetention.c
```diff
@@ -364,6 +374,7 @@ int32_t tsdbRetention(void *arg) {
   // clear resources
   tsdbTFileSetClear(&rtner.fset);
   TARRAY2_DESTROY(&rtner.fopArr, NULL);
+  (void)tsdbRemoveRetentionMonitorTask(pTsdb, &rtnArg->taskid);
```
localvar, line 424: this should be protected by a lock, or concurrent calls to `tsdbRetention` may result in race condition.
calls to other retention monitor functions may also have similiar issues.

source/dnode/vnode/src/tsdb/tsdbFS2.c
```diff
@@ -1405,7 +1408,10 @@ int32_t tsdbFileSetGetEntryField(struct SFileSetReader *pReader, const char *fie
 
   fieldName = "should_compact";
   if (strncmp(field, fieldName, strlen(fieldName) + 1) == 0) {
-    *(char *)value = tsdbShouldCompact(pReader->pFileSet, pReader->pTsdb->pVnode->config.vgId);
+    *(char *)value = false;
+#ifdef TD_ENTERPRISE
+    *(char *)value = tsdbShouldCompact(pReader->pFileSet, pReader->pTsdb->pVnode->config.vgId, TSDB_OPTR_NORMAL);
+#endif
```
localvar:
```suggestion
    *(bool *)value = false;
#ifdef TD_ENTERPRISE
    *(bool *)value = tsdbShouldCompact(pReader->pFileSet, pReader->pTsdb->pVnode->config.vgId, TSDB_OPTR_NORMAL);
#endif
```

source/dnode/vnode/src/vnd/vnodeCfg.c
```diff
@@ -342,6 +343,8 @@ int vnodeDecodeConfig(const SJson *pJson, void *pObj) {
   if (code) return code;
   tjsonGetNumberValue(pJson, "vndStats.ntimeseries", pCfg->vndStats.numOfNTimeSeries, code);
   if (code) return code;
+  tjsonGetNumberValue(pJson, "vndStats.rsmas", pCfg->vndStats.numOfRSMAs, code);
+  if (code) return code;
```
localvar, line 347: in case of upgrading an existing instance, will it always return here because there's no `vndStats.rsmas`in the config?

kailixu: code is 0 if no "vndStats.rsmas" exists.

include/common/tmsg.h
kailixu: They'e used for compact and retention, mean for the origin of the operations, both in mnode and vnode.

source/dnode/vnode/src/tsdb/tsdbFS2.c
kailixu: OK

source/dnode/vnode/src/tsdb/tsdbRetention.c
kailixu: 1. Mutex lock exists inside of most of the functions. 2. For tsdbAddRetentionMonitorTask, the lock is added outside in tsdbAsyncRetention/tsdbAsyncRetentionImpl.

PR #33173, opened 2025-10-01
Include default libtaospyudf.so location to LD_LIBRARY_PATH

alexey-inno:
# Description

Hi, I propose adding default `libtaospyudf.so` location to `LD_LIBRARY_PATH`

This is not a breaking issue, but certainly another thing that's easy to miss. I'm not sure if "docs update" to warn about this would be a good solution, but perhaps there is also a place to update docs with a note, since not everybody runs TDEngine is a container

I'd assume it's better to make things run seamlessly, without too much tinkering, especially for a docker image

# Checklist

- [X] Is there no significant decrease in test coverage?


alexey-inno:
Not sure how to run Docker-related tests locally. Installing locally without Docker failed

I can try to debug it, but not sure how much impact this would do since we'd only need docker-related tests to pass

PR #33550, opened 2025-11-12
fix: variable-length type fields distinguish between null and empty values for taosdump

YamingPei:
# Description

Variable-length type fields distinguish between null and empty values for taosdump
Jira: https://jira.taosdata.com:18080/browse/TD-38549

# Checklist

- [x] Is the user manual updated?


zitsen:
Failed cases are all stream related
image

PR #33607, opened 2025-11-15
fix(taosBenchmark): fix disorder data count verification logic and support setting tolerance

YamingPei:
# Description

Fix disorder data count verification logic and support setting tolerance

Jira: https://jira.taosdata.com:18080/browse/TD-38574

# Checklist

### DIFF
--- a/test/new_test_framework/utils/sql.py
+++ b/test/new_test_framework/utils/sql.py
@@ -1620,14 +1620,12 @@ class TDSql:
                 return True
 
             elif isinstance(data, int) and tolerance != 0.0:
-                if tolerance < 0:
-                    lower = data * (1 + tolerance)
-                    upper = data
-                else:
-                    lower = data
-                    upper = data * (1 + tolerance)
+                bound1 = data
+                bound2 = data * (1 + tolerance)
+                lower = min(bound1, bound2)
+                upper = max(bound1, bound2)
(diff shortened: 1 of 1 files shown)


YamingPei:
image
失败用例与本PR无关。

test/new_test_framework/utils/sql.py
```diff
@@ -1618,6 +1618,40 @@ def checkData(self, row, col, data, show=False, exit=True, dbPrecision=""):
                     else:
                         return False
                 return True
+
+            elif isinstance(data, int) and tolerance != 0.0:
+                if tolerance < 0:
+                    lower = data * (1 + tolerance)
+                    upper = data
```
sheyanjie-qq: can we make sure data > 0 ?

YamingPei: No need. The data can be any integer, and the tolerance can be positive or negative. As long as the data passed in by the business and the verification logic are consistent, it is fine.

YamingPei: Under the current test case, when the tolerance parameter is passed in, the data is always non-negative.
When the data is negative, the upper and lower limits of the data range are reversed. This has been corrected.

YamingPei:
image
失败case与本PR无关。

Issue #34544, opened 2026-02-09
taosd SEGV in addTagPseudoColumnData when TMQ polls while streams write to same vgroup (macOS ARM64)


brunohaid:
AI summary but human verified:

We're running TDengine 3.4.0.2 Community on macOS ARM64 (Apple Silicon) for local development. We have 10 streams processing collectd metrics and a TMQ consumer subscribed to the raw collectd supertables. taosd consistently crashes with a null pointer dereference after 30-60 seconds of TMQ polling.

The crash is in `addTagPseudoColumnData` — it gets a NULL pointer for a subtable's tag data and passes it straight to `memmove` via `doCopyNItems`.

`addTagPseudoColumnData` looks up the tag value (like host) for a subtable that a stream is concurrently writing to, gets back a NULL pointer because the metadata isn't fully initialized yet, and passes it to memmove which dereferences it.

lldb:

```
* thread #91, name = 'vnode-query', stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: libsystem_platform.dylib`_platform_memmove + 168
    frame #1: taosd`doCopyNItems + 260
    frame #2: taosd`addTagPseudoColumnData + 1280
    frame #3: taosd`doQueueScanNext + 1204
    frame #4: taosd`getNextBlockFromDownstreamImpl + 312
    frame #5: taosd`getNextBlockFromDownstream + 32
    frame #6: taosd`doProjectOperation + 260
    frame #7: taosd`qExecTask + 576
    frame #8: taosd`getDataBlock + 132
    frame #9: taosd`tqScanData + 164
    frame #10: taosd`tqExtractDataForMq + 688
    frame #11: taosd`tqProcessPollReq + 984
    frame #12: taosd`vnodeProcessQueryMsg + 344
    frame #13: taosd`vmProcessQueryQueue + 216
    frame #14: taosd`tQueryAutoQWorkerThreadFp + 632
```

At the time of the crash, a `snode-stream-runner` thread was actively running a multi-table JOIN stream (`core_stream`) that writes to an output supertable in the same vgroup. That thread's backtrace shows 5 levels of `mInnerJoinDo` → `mJoinMainProcess` → `doProjectOperation` → `streamExecuteTask`. So we have concurrent stream writes and TMQ reads touching the same vgroup's subtable metadata.

Our setup:
- 6 raw collectd supertables (cpuavg_value, cpumax_value, memory_value, disk_value, if_0, if_1)
[...]

PR #34890, opened 2026-03-22
fix(taosd): follow-up guards for tag pseudo-column crash path

brunohaid:
Follow-up hardening for issue #34544: https://github.com/taosdata/TDengine/issues/34544

After testing commit 6db0339 on macOS arm64 under TMQ + streams pressure, we still observed a crash in the same path:
addTagPseudoColumnData -> colDataSetNItems -> doCopyNItems -> memmove

What this PR changes (surgical):
1) source/common/src/tdatablock.c
- doCopyNItems(...): early return when numOfRows <= 0

2) source/libs/executor/src/scanoperator.c
- createTableCacheVal(...): null-check pMetaReader->me.ctbEntry.pTags before dereference/copy
- addTagPseudoColumnData(...): only call extractTagVal(...) when val.pTags != NULL

Why:
- Prevents zero-row copy crash mode
- Prevents null tag-buffer dereference in child-table metadata edge cases
- Keeps normal behavior unchanged

Scope:
- Minimal defensive guards only
- No architectural changes

Related to #34544

### DIFF
--- a/source/common/src/tdatablock.c
+++ b/source/common/src/tdatablock.c
@@ -274,6 +274,10 @@ static int32_t colDataReserve(SColumnInfoData* pColumnInfoData, size_t newSize) 
 
 int32_t doCopyNItems(struct SColumnInfoData* pColumnInfoData, int32_t currentRow, const char* pData,
                             int32_t itemLen, int32_t numOfRows, bool trimValue) {
+  if (numOfRows <= 0) {
+    return TSDB_CODE_SUCCESS;
+  }
+
   if (pColumnInfoData->info.bytes < itemLen) {
     uWarn("column/tag actual data len %d is bigger than schema len %d, trim it:%d", itemLen,
           pColumnInfoData->info.bytes, trimValue);
--- a/source/libs/executor/src/scanoperator.c
+++ b/source/libs/executor/src/scanoperator.c
@@ -557,9 +557,11 @@ static int32_t createTableCacheVal(const SMetaReader* pMetaReader, STableCachedV
   // only child table has tag value
   if (pMetaReader->me.type == TSDB_CHILD_TABLE || pMetaReader->me.type == TSDB_VIRTUAL_CHILD_TABLE) {
     STag* pTag = (STag*)pMetaReader->me.ctbEntry.pTags;
-    pVal->pTags = taosMemoryMalloc(pTag->len);
-    QUERY_CHECK_NULL(pVal->pTags, code, lino, _end, terrno);
-    memcpy(pVal->pTags, pTag, pTag->len);
+    if (pTag != NULL) {
+      pVal->pTags = taosMemoryMalloc(pTag->len);
+      QUERY_CHECK_NULL(pVal->pTags, code, lino, _end, terrno);
+      memcpy(pVal->pTags, pTag, pTag->len);
+    }
   }
 
   (*ppResVal) = pVal;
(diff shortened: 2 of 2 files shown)


guanshengliang (approved):

JinqingKuang:
## Code Review Summary

Scope: +13 / -4 across 2 files. The three defensive guards are well-aimed at the `addTagPseudoColumnData → colDataSetNItems → doCopyNItems → memmove` crash path from #34544. The two findings below are **maintainability-only** — no correctness, security, performance, or rule-scope problems were validated.

This review is static analysis against `base..HEAD`; no CI/build/lint/test was run.

### 1. Guard #3 (`extractTagVal` NULL-check) is redundant given the current contract, while four sibling call sites remain unguarded

`source/libs/executor/src/scanoperator.c:702-705` now wraps the `extractTagVal` call in `if (val.pTags != NULL)`. The only registered implementation, `metaGetTableTagVal` (`source/dnode/vnode/src/meta/metaQuery.c:1216-1228`), already tolerates a NULL `pTag`:

- JSON branch: returns the NULL pointer directly.
- Non-JSON branch: delegates to `tTagGet`, which null-checks `pTag` at `source/common/src/tdataformat.c:2874-2877` and returns `false` → `NULL`.

So after guard #2, a cached `val.pTags == NULL` is already safe to pass into `extractTagVal`. Guard #3 adds no behavior change here.

Meanwhile, the same `extractTagVal(*ctbEntry.pTags, …)` pattern is **unguarded** at:

- `source/libs/executor/src/scanoperator.c:3462`
- `source/dnode/vnode/src/vnd/vnodeStream.c:1349`
- `source/dnode/vnode/src/vnd/vnodeStream.c:3938`
- `source/libs/executor/src/executil.c:504`

If the intended mental model is "`pTags` may be NULL, guard before `extractTagVal`" (as guard #3 implies), those four sites are inconsistent. If the model is "`extractTagVal` handles NULL", guard #3 is dead defensive code that future readers will likely copy.

**Suggested direction:** pick one rule and apply it uniformly — either drop guard #3 and rely on `extractTagVal`'s existing NULL tolerance, or keep it and apply the same guard at the four sibling sites (or push the check into `metaGetTableTagVal`).

### 2. `createTableCacheVal` silently changes its post-condition — successful return can now leave `pVal->pTags == NULL`, but the contract is undocumented

Before this PR, a successful `createTableCacheVal` for a child / virtual-child table guaranteed `pVal->pTags != NULL` (unconditional `taosMemoryMalloc` + `memcpy`). After this PR (`source/libs/executor/src/scanoperator.c:557-563`), success can leave `pVal->pTags == NULL`.

That cached value flows through `taosLRUCache` and is read at `scanoperator.c:668-672` via `val = *pValTmp;`. Today's only consumer is the function changed in this PR, but `STableCachedVal` is a shared cache entry. The next caller to copy the pattern `val = *pValTmp; … extractTagVal(val.pTags, …)` into a new path will be silently exposed to whatever the relevant `extractTagVal` impl does on NULL — and there is no comment on the struct or the new `if (pTag != NULL)` branch signaling this contract change.

`freeTableCachedVal` already tolerates `taosMemoryFree(NULL)`, so no leak — purely a documentation gap.

**Suggested direction:** add a one-line comment at the new `if (pTag != NULL)` block (e.g. `// pTags may be NULL when ctbEntry has no tag buffer; consumers must null-check`) and/or on `STableCachedVal`. No code change required beyond the comment.

### Notes on the other two changes

- `doCopyNItems` early return on `numOfRows <= 0` — real fix; previously the unconditional first `memcpy` and `log(numOfRows)/log(2)` ran on a zero-row path.
- `createTableCacheVal` `pTag` null-check — real fix; the original code dereferenced `pTag->len` and `memcpy(pTag)` unconditionally.

Both are surgical and correctly targeted at the #34544 crash mode.

### Review boundaries

- Static analysis only; no CI/build/lint/test was executed.
- Both findings are cross-cutting (contract / consistency), so no inline comments — summary only.

`Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>`

PR #34910, opened 2026-03-24
fix: improve GCC compatibility for util build

tomchon:
fix : https://project.feishu.cn/taosdata_td/defect/detail/6927859716
- guard taosFloatHash no_sanitize attribute by compiler support
- add -Wno-stringop-overread only when the compiler supports it
- avoid build failures on older GCC with -Werror enabled


dapan1121 (commented):
LGTM

guanshengliang (approved):

guanshengliang (approved):

PR #34954, opened 2026-03-27
enh:stream nodelay create out table

Pengrongkun:
# Description

# Issue(s)

- Close/close/Fix/fix/Resolve/resolve: Issue Link

# Checklist


JinqingKuang:
### Code review

Found 3 issues:

1. **Undo action `SVDropTbReq.name` is never set, causing transaction rollback to fail** — In `mndStreamCreateOutTable`, the drop table undo action initializes `dropReq` with only `uid`, `igNotExists`, and `isVirtual`, but leaves `name` as NULL. When the undo action executes, `metaCheckDropTableReq` returns `TSDB_CODE_INVALID_MSG` for a NULL name. The action's `acceptableCode` is only `TSDB_CODE_TDB_TABLE_NOT_EXIST`, so this mismatch causes the undo action to fail with an unacceptable error code, blocking transaction rollback and potentially leaving orphaned tables in the vnode.

https://github.com/taosdata/TDengine/blob/f12fedde3726969e70400a0e1cf5fa000afc67e1/source/dnode/mnode/impl/src/mndStream.c#L520-L526

2. **`checks3()` asserts `getRows() == 1` but then accesses row index 1** — The lambda in `checks3` checks `tdSql.getRows() == 1` (expecting a single row), then calls `tdSql.compareData(1, 0, "out3_tb2")` which accesses out-of-bounds row index 1. Either the row count assertion should be 2, or the second `compareData` call is wrong.

https://github.com/taosdata/TDengine/blob/f12fedde3726969e70400a0e1cf5fa000afc67e1/test/cases/18-StreamProcessing/99-Others/auto_create_output_table.py#L117-L133

3. **New test file `auto_create_output_table.py` not registered in `community/test/ci/cases.task`** — The project CLAUDE.md requires: "新增系统测试用例文件后，必须在 `community/test/ci/cases.task` 中注册对应条目". This PR adds the file but does not update `cases.task`, so the new test will never run in CI.

https://github.com/taosdata/TDengine/blob/f12fedde3726969e70400a0e1cf5fa000afc67e1/test/cases/18-StreamProcessing/99-Others/auto_create_output_table.py#L1-L5

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #35044, opened 2026-04-02
fix(stream): tolerate existing output table on submit;fix stream out-column timestamp precision checks

Pengrongkun:
…ut-column precision checks

# Description

Fixes stream writes to pre-created normal output tables when NODELAY_CREATE_SUBTABLE is used: the vnode can return TSDB_CODE_TDB_TABLE_ALREADY_EXIST for the embedded create step, which must be treated as success on the client. Also adjusts CREATE STREAM validation so timestamp precision is checked against the table time precision, and decimal columns still require matching precision; other types are no longer rejected solely on column-level precision mismatch.

# Issue(s)

https://project.feishu.cn/taosdata_td/defect/detail/6931571628

# Checklist


test/cases/18-StreamProcessing/99-Others/test_auto_create_output_table.py
```diff
@@ -245,6 +275,28 @@ def insertDataAndCheck(self):
             and tdSql.compareData(0, 1, "10001")
         )
 
+        tdSql.execute("insert into tb1 values ('2025-01-01 00:00:02', 10003);")
+        time.sleep(5)
+
```
JinqingKuang, line 280: `tdSql.checkResultsByFunc()` 已经会轮询重试最多 300 次，所以这里再写死一个 `time.sleep(5)` 只会增加测试耗时，不会让断言更可靠。更干净的做法是删掉这个 sleep，直接依赖现有轮询；如果确实要调等待行为，就显式传 `delay`/`retry` 给 `checkResultsByFunc()`。

JinqingKuang:
代码审查更正：

当前仍然成立的问题只有 1 个：

1. 新测试在结果校验前加了固定的 `time.sleep(5)`。但 `tdSql.checkResultsByFunc()` 本身已经带轮询重试，这个 sleep 只会增加测试耗时，不会提升正确性。

PR #35388, opened 2026-06-12
fix(util): reject oversize length in binary message decoders

dxbjavid:
**Wrapping length check in the binary decoders**
The bounds check in `tDecodeBinaryWithSize` reads as `pCoder->pos + size > pCoder->size`, but `pos` and `size` are both `uint32_t` so the addition is done in 32 bits and wraps. `size` arrives from a varint in the message via `tDecodeBinary`, so a length close to `UINT32_MAX` (encoded as `ff ff ff ff 0f`) makes the sum wrap below `size` and the check passes. The decoder then returns a pointer with that oversize length and advances `pos` past the end of the buffer, so the next read runs out of bounds. I switched the four binary decoders that take a message-supplied length to the non-wrapping form `size > pCoder->size - pCoder->pos`, which is safe because `pos` never exceeds `size`.

### DIFF
--- a/include/util/tencode.h
+++ b/include/util/tencode.h
@@ -412,7 +412,7 @@ static int32_t tDecodeBool(SDecoder* pCoder, bool* val) {
 }
 
 static FORCE_INLINE int32_t tDecodeBinaryWithSize(SDecoder* pCoder, uint32_t size, uint8_t** val) {
-  if (pCoder->pos + size > pCoder->size) {
+  if (size > pCoder->size - pCoder->pos) {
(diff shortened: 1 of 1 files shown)


dxbjavid:
Hi maintainers,

Just a friendly follow-up on this PR. When you have a chance, could you please provide an update on its review status?

Happy to make any additional changes if needed.

Thank you!

dxbjavid:
CLA done!

dxbjavid:
Any update?

dxbjavid:
Closing this pull request due to inactivity

PR #35405, opened 2026-07-16
fix(cmake): preserve Windows build OS metadata

VectorPeak:
# Description

Preserve the platform-derived build CPU/OS metadata when `options.cmake` registers `BUILD_VER_CPUTYPE` and `BUILD_VER_OSTYPE` as cache overrides.

## What Problem This Solves

The top-level CMake load order is:

```text
platform.cmake -> options.cmake -> version.cmake
```

`platform.cmake` computes platform metadata such as `BUILD_VER_CPUTYPE=x64` and `BUILD_VER_OSTYPE=Windows`. `options.cmake` then declared the same names as empty cache strings, and `version.cmake` could fall back to generic defaults when those values were empty.

That could make a Windows configure report contradictory metadata:

```text
Current system: Windows
cpu: x64
os: Linux
```

The cache entries still need to support user overrides such as `-DBUILD_VER_OSTYPE=CustomOS` and `-DBUILD_VER_CPUTYPE=arm64`, so the fix cannot simply hardcode platform values later in `version.cmake`.

## Changes

`options.cmake` now handles both build metadata cache variables in two cases:

```diff
-set(BUILD_VER_CPUTYPE "" CACHE STRING "CPU type override")
-set(BUILD_VER_OSTYPE  "" CACHE STRING "OS type override")
+if(DEFINED CACHE{BUILD_VER_CPUTYPE} AND NOT "$CACHE{BUILD_VER_CPUTYPE}" STREQUAL "")
+  set(BUILD_VER_CPUTYPE "$CACHE{BUILD_VER_CPUTYPE}")
+else()
+  set(BUILD_VER_CPUTYPE "${BUILD_VER_CPUTYPE}" CACHE STRING "CPU type override" FORCE)
+endif()
+if(DEFINED CACHE{BUILD_VER_OSTYPE} AND NOT "$CACHE{BUILD_VER_OSTYPE}" STREQUAL "")
+  set(BUILD_VER_OSTYPE "$CACHE{BUILD_VER_OSTYPE}")
+else()
+  set(BUILD_VER_OSTYPE "${BUILD_VER_OSTYPE}" CACHE STRING "OS type override" FORCE)
+endif()
```

- If a non-empty cache value already exists, it is treated as the user's explicit override and copied back into the normal variable.
- Otherwise, the platform values computed by `platform.cmake` are used as the cache defaults.

## Evidence

A minimal CMake script that simulates the relevant variable/cache sequence now preserves platform defaults and explicit overrides:

```text
Windows x64 default:
-- cpu: x64
-- os: Windows
Linux arm64 default:
-- cpu: arm64
-- os: Linux
Overrides:
-- cpu: arm64
-- os: CustomOS
```

A real Windows configure reached the version block with corrected defaults:

```text
cmake -S . -B %TEMP%\tdengine-cmake-buildver-default -G "Visual Studio 17 2022" -A x64 -DBUILD_CONTRIB=OFF -DBUILD_ROCKSDB=OFF -DBUILD_TOOLS=OFF -DBUILD_TEST=OFF -DBUILD_USE_PUBLIC_DEPS=ON
-- Current system: Windows
-- cpu:        x64
-- os:         Windows
```

The same configure with explicit overrides preserved the user values:

```text
cmake -S . -B %TEMP%\tdengine-cmake-buildver-override -G "Visual Studio 17 2022" -A x64 -DBUILD_CONTRIB=OFF -DBUILD_ROCKSDB=OFF -DBUILD_TOOLS=OFF -DBUILD_TEST=OFF -DBUILD_USE_PUBLIC_DEPS=ON -DBUILD_VER_OSTYPE=CustomOS -DBUILD_VER_CPUTYPE=arm64
-- Current system: Windows
-- cpu:        arm64
-- os:         CustomOS
```

Both configure attempts later stopped in the external dependency stage because the local checkout does not have the cached prebuilt RocksDB artifact; the corrected `cpu:` and `os:` metadata is printed before that unrelated stop.

## Possible call chain / impact

```text
CMakeLists.txt
  -> include(cmake/platform.cmake)  # sets BUILD_VER_CPUTYPE / BUILD_VER_OSTYPE from platform
  -> include(cmake/options.cmake)   # registers cache overrides without erasing defaults
  -> include(cmake/version.cmake)   # prints and stores build metadata
```

This change is limited to build metadata initialization. It does not alter platform detection, compiler flags, dependency selection, or user-provided non-empty `BUILD_VER_CPUTYPE` / `BUILD_VER_OSTYPE` overrides.

# Checklist


VectorPeak:
Addressed the reviewer suggestion by applying the same cache/normal-variable preservation pattern to `BUILD_VER_CPUTYPE` as well as `BUILD_VER_OSTYPE`.

Validation run locally:

```text
Windows x64 default:
-- cpu: x64
-- os: Windows
Linux arm64 default:
-- cpu: arm64
-- os: Linux
Overrides:
-- cpu: arm64
-- os: CustomOS
```

A real Windows configure also reached the version block with `cpu: x64` / `os: Windows`, and with explicit overrides preserved `cpu: arm64` / `os: CustomOS`. Both configure attempts later stopped at the unrelated cached RocksDB artifact check, after the build metadata had already been printed.

guanshengliang:

- Internal MR: https://git.tdengine.net/rd-public/tsdb/-/merge_requests/1954

PR #35406, opened 2026-07-17
fix(function): reject SM4 functions in community edition

### LINKED ISSUE #35349: Logic Flaw: SM4_ENCRYPT/DECRYPT functions return plaintext without warning in Community Edition
**Bug Description**
In the Community Edition build, the SM4 cryptographic routines are stubbed out using a `memcpy` operation, but the system processes the functions with a silent success status (`Query OK`). This creates a critical logical flaw and a false sense of security, as users are led to believe their data is encrypted when it is actually processed and stored as raw plaintext.

**To Reproduce**
1. Open the interactive client terminal by running `./taos`.
2. Select or use any database context (`USE test_vuln;`).
3. Execute the SM4 encryption query followed by the AES encryption query to see the plaintext leak vs expected behavior:

```sql
taos> SELECT SM4_ENCRYPT('HelloGhost12345', 'my_sm4_key_16_ch') FROM secure_data;
 sm4_encrypt('HelloGhost12345', 'my_sm4_key_16_ch') |
=====================================================
 HelloGhost12345                                    |
Query OK, 1 row(s) in set (0.066803s)

taos> SELECT AES_ENCRYPT('HelloGhost12345', 'my_aes_key_16_ch') FROM secure_data;
 aes_encrypt('HelloGhost12345', 'my_aes_key_16_ch') |
=====================================================
                                                    |
Query OK, 1 row(s) in set (0.011350s)

```

4. Notice that the SM4 output returns the raw, unencrypted plaintext string `HelloGhost12345` while `Query OK` is advertised.

**Expected Behavior**
If the SM4 algorithm is restricted or unimplemented in the Community Edition, the engine should explicitly fail and throw a proper error code (e.g., `Function not supported in this edition`) instead of failing open, silently copying plaintext, and misleading the operator.

**Environment:**

* **OS:** Debian
* **TDengine Version:** 3.4.1.6

**Additional Context**
The root cause lies in `source/libs/crypt/src/crypt.c`. When `TD_ENTERPRISE` is not defined, the core routines fall back to `Builtin_CBC_EncryptImpl`, which performs a direct data copy:

```c
int32_t Builtin_CBC_EncryptImpl(SCryptOpts *opts) {
    memcpy(opts->result, opts->source, opts->len);
    return opts->len;
}

```

While commercial gating is understandable, performing a silent `memcpy` without a runtime warning propagates unencrypted data into architectural call sites (like WAL logs and local configurations) while advertising them as secure.

bestdo77:
Fixes #35349.

## Problem

In the community edition, `SM4_ENCRYPT`/`SM4_DECRYPT` report success but return the plaintext unchanged: the community build ships a copy-through SM4 implementation (kept for internal storage encryption such as WAL/checkpoint), and the SQL layer did not reject these functions. Users therefore believe their data is encrypted while it is stored in plaintext.

## Changes

Reject the two SM4 SQL functions in community builds with `TSDB_CODE_OPS_NOT_SUPPORT` ("only supported in Enterprise edition"), at both entry points:

- `source/libs/function/src/builtins.c`: `translateOutSm4`/`translateOutSm4De` return the error during function translation — covers all query paths (`SELECT`, `WHERE`, `INSERT ... SELECT`, ...).
- `source/libs/parser/src/parInsertSql.c`: `parseBinary` rejects `TK_SM4_ENCRYPT`/`TK_SM4_DECRYPT` — covers the `INSERT ... VALUES(sm4_encrypt(...))` constant-folding fast path, which bypasses function translation.

Enterprise builds (`TD_ENTERPRISE`/`TD_ASTRA`) are unchanged; the guard macro follows the existing convention in `source/libs/crypt/src/crypt.c`.

## Tests

- `test/cases/11-Functions/01-Scalar/test_scalar_crypto.py`: probe SM4 availability, assert the enterprise-only error on SELECT/INSERT encrypt/decrypt paths in community edition; original positive cases now run only where SM4 is supported.
- `test/cases/11-Functions/09-NoFrom/test_fun_no_from_all.py`: SM4 no-FROM cases moved to a separate list and routed to positive/negative assertions based on a runtime probe, so the case passes on both editions. Also pass the test client config dir (`-c`) to the `taos` CLI invocations.

## Verification (community build)

- Incremental build and parser unit tests pass.
- Both Python cases pass.
- Live node: `SELECT sm4_encrypt/decrypt(...)`, `INSERT ... VALUES(sm4_encrypt/decrypt(...))` all fail with `0x80000100 only supported in Enterprise edition`; AES functions unaffected; rejected SM4 inserts write no plaintext.
- Environment: community build `3.4.1.13.alpha.community` (base `main@df445db5dc`), Linux x86_64. Enterprise edition was not run locally; it is unaffected by construction since the rejection is compiled out when `TD_ENTERPRISE`/`TD_ASTRA` is defined.

### DIFF
--- a/source/libs/function/src/builtins.c
+++ b/source/libs/function/src/builtins.c
@@ -49,6 +49,13 @@ static int32_t invaildFuncParaValueErrMsg(char* pErrBuf, int32_t len, const char
   return buildFuncErrMsg(pErrBuf, len, TSDB_CODE_FUNC_FUNTION_PARA_VALUE, "Invalid parameter value : %s", pFuncName);
 }
 
+#if !defined(TD_ENTERPRISE) && !defined(TD_ASTRA)
+static int32_t enterpriseOnlyFuncErrMsg(char* pErrBuf, int32_t len, const char* pFuncName) {
+  return buildFuncErrMsg(pErrBuf, len, TSDB_CODE_OPS_NOT_SUPPORT, "%s is only supported in Enterprise edition",
+                         pFuncName);
+}
+#endif
+
 static int32_t validateTimeUnitParam(uint8_t dbPrec, const SValueNode* pVal) {
   if (!IS_DURATION_VAL(pVal->flag)) {
     return TSDB_CODE_FUNC_TIME_UNIT_INVALID;
@@ -1793,6 +1800,10 @@ static int32_t translateOutAesDe(SFunctionNode* pFunc, char* pErrBuf, int32_t le
 }
 
 static int32_t translateOutSm4(SFunctionNode* pFunc, char* pErrBuf, int32_t len) {
+#if !defined(TD_ENTERPRISE) && !defined(TD_ASTRA)
+  return enterpriseOnlyFuncErrMsg(pErrBuf, len, pFunc->functionName);
+#endif
+
   FUNC_ERR_RET(validateParam(pFunc, pErrBuf, len));
 
   uint8_t orgType = getSDataTypeFromNode(nodesListGetNode(pFunc->pParameterList, 0))->type;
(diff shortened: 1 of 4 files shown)


guanshengliang:

- Internal MR: https://git.tdengine.net/rd-public/tsdb/-/merge_requests/1953

PR #35408, opened 2026-07-17
fix(function): bound group_concat result and clamp stream out columns

### LINKED ISSUE #34504: 3.4.0.2版本中在计数窗口流计算中使用group_concat聚合函数会引发宕机和报错
**Bug Description**
3.4.0.2版本中在计数窗口流计算中使用group_concat聚合函数会引发宕机和报错

**Environment (please complete the following information):**
 - OS: Linux
 - TDengine Version  3.4.2

**Additional Context**
建表脚本如下：
taos> CREATE STABLE IF NOT EXISTS s_ptr_data (
    >     ts TIMESTAMP,
    >     test_value double,
    >     test_value_str VARCHAR(16),
    >     test_flag INT,
    >     hi_limit double,
    >     lo_limit double
    > ) TAGS (
    >     factory VARCHAR(32),
    >     fileid VARCHAR(64),
    >     testnum INT,
    >     testtxt VARCHAR(256),
    >     headnum INT,
    >     sitenum INT,
    >     productid VARCHAR(64),
    >     lotid VARCHAR(64),
    >     sublotid VARCHAR(64),
    >     jobname VARCHAR(128),
    >     flowid VARCHAR(32),
    >     datatype VARCHAR(10),
    >     nodename VARCHAR(32),
    >     dutindex INT
    > );
Create OK, 0 row(s) affected (0.015924s)

创建流计算脚本如下
taos> 
taos> CREATE STREAM s1 COUNT_WINDOW(10)
    > FROM s_ptr_data PARTITION BY tbname
    > INTO output
    > AS
    > SELECT LAST(ts) AS ts, AVG(test_value) AS avg_test_value, GROUP_CONCAT(test_value_str, ',') AS test_value_array
    > FROM s_ptr_data;

DB error: Row length exceeds max length 65531 [0x8000263E] (0.029885s)

当s_ptr_data 表写入数据后，会引发所有数据库的连接超时以及宕机
目前观察下来，docker，window端都会崩溃

bestdo77:
Fixes #34504.

group_concat has two problems:

1. The result is accumulated into a fixed 65519-byte buffer with no bounds check (`gconcatHelper` in builtinsimpl.c). Once a group's payload grows past that, the memcpy corrupts the heap and taosd aborts (`malloc(): corrupted top size`, signal 6). I reproduced this on current main by running group_concat over ~6k rows (~102 KB payload) — the server dies immediately.

2. `translateGroupConcat` always estimates the result width as 65515, so a stream output row containing any other column exceeds the 65531 limit and `CREATE STREAM` fails with `Row length exceeds max length 65531 [0x8000263E]` — the error from the issue.

What this PR does:

- `gconcatHelper` now checks the length before every append and returns "Value too long for column/tag" when the payload would exceed the limit. The query fails, the server stays up. A `cleanupFunc` is registered for group_concat and all buffer frees are idempotent, so the error path releases the accumulation buffer instead of leaking it.
- When the auto-derived stream output row is too wide, group_concat output columns are clamped to what fits (`createStreamReqSetDefaultOutCols`). The stream in the issue can be created now. If a real value is still longer than the clamped column, the write is rejected with a normal error instead of crashing.

Reproduced/tested with:

```sql
-- 600+ rows of 120+ byte values: before = server aborts, after = clean error
select group_concat(v, ',') from t;

-- before = 0x8000263E row too long, after = created
create stream s1 count_window(10) from s_ptr_data partition by tbname into output as
  select last(ts) as ts, avg(test_value) as avg_test_value, group_concat(test_value_str, ',') as test_value_array
  from s_ptr_data;
```

Added `test_group_concat_overflow` to `test_agg_gconcat.py` covering both cases.

One note for the reporter: the SQL in the issue has no window constraint (`ts >= _twstart and ts <= _twend`), so each trigger scans the whole table — that is how the calc query behaves without it; the stream cases in this repo always include it.

Tested on Debian 10 x86_64, community build 3.4.1.13.alpha based on main@df445db5:

- parserTest stream suite passes (30/30)
- test_agg_gconcat.py 3/3, test_agg_smoking.py 6/6
- on a live node the overflow query returns 0x80002653 in ~10 ms and the server keeps serving; the stream above is created (output row 65533 -> 65531 after the clamp)

Performance impact should be negligible: one integer comparison per appended value in group_concat, and the clamp runs once per stream creation only when the row is oversized. No other code paths change. The only behavior change is that an oversized group_concat now returns an error instead of crashing the server.

### DIFF
--- a/source/libs/function/inc/builtinsimpl.h
+++ b/source/libs/function/inc/builtinsimpl.h
@@ -85,6 +85,7 @@ bool    gconcatGetFuncEnv(SFunctionNode* pFunc, SFuncExecEnv* pEnv);
 int32_t gconcatFunctionSetup(SqlFunctionCtx* pCtx, SResultRowEntryInfo* pResultInfo);
 int32_t gconcatFunction(SqlFunctionCtx* pCtx);
 int32_t gconcatFinalize(SqlFunctionCtx* pCtx, SSDataBlock* pBlock);
+void    gconcatFunctionCleanupExt(SqlFunctionCtx* pCtx);
 
 bool    getLeastSQRFuncEnv(struct SFunctionNode* pFunc, SFuncExecEnv* pEnv);
 int32_t leastSQRFunctionSetup(SqlFunctionCtx* pCtx, SResultRowEntryInfo* pResultInfo);
--- a/source/libs/function/src/builtins.c
+++ b/source/libs/function/src/builtins.c
@@ -6878,6 +6878,7 @@ const SBuiltinFuncDefinition funcMgtBuiltins[] = {
     .initFunc     = gconcatFunctionSetup,
     .processFunc  = gconcatFunction,
     .finalizeFunc = gconcatFinalize,
+    .cleanupFunc  = gconcatFunctionCleanupExt,
     /*
     .combineFunc  = gconcatCombine, // stream
     .pStateFunc = "group_concat",   // tsma
--- a/source/libs/function/src/builtinsimpl.c
+++ b/source/libs/function/src/builtinsimpl.c
@@ -1585,6 +1585,8 @@ static int32_t gconcatHelper(const char* input, char* output, bool hasNchar, int
   if (input == NULL) {
     return TSDB_CODE_SUCCESS;
   }
+  // the result buffer is allocated with TSDB_MAX_FIELD_LEN bytes (varstr header included), and the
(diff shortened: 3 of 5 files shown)


guanshengliang:

- Internal MR: https://git.tdengine.net/rd-public/tsdb/-/merge_requests/1952

